PR #1, opened 2017-06-08
add ColoredSmtpAppender. allow user send colorful log email.

dpsenner:
Opening this branch as a pull request to give us a place for discussion.


dpsenner:
After reading this patch I think we should provide a base class that has a generic way to have customized mail messages. All the customizations (i.e. for this code the log level to color conversions) can then happen in a class that derives from that appender. This gives users all the possibilities there are to have custom mail messages while reducing the effort to write such an appender. In log4net core I do not see room for such things unless we provide a templating engine.

PR #2, opened 2017-06-08
First version of ProcMonAppender

dpsenner:
Reopening this pull request, please discuss.


dpsenner:
I wonder whether we would like to have this windows specific appender in core. Do we need a separate assembly for appenders that are second class citizens?

zippy1981:
This appender silently fails on windows if procmon isn't running. If I made the same thing happen on linux (probably have to do some try/catch during the interop calls) Can we include it.

Also, whats the minum .net framework your currently targeting? I think new versions of .net or at least .NET core will let me do this without P/Invoke.

zippy1981:
@dpsenner Waiting for your response to my comments.

PR #3, opened 2017-06-08
Support for types of non-parameterless constructors in XmlConfigurator.

dpsenner:
The constructor parameters are taken from the corresponding values of
the child nodes.

You can add for instance

<standardunit type="Amazon.CloudWatch.StandardUnit">
<value value="Kilobytes"/>
</standardunit>

where StandardUnit has only one constructor that has a string parameter
named "value".

### DIFF
--- a/src/log4net/Repository/Hierarchy/XmlHierarchyConfigurator.cs
+++ b/src/log4net/Repository/Hierarchy/XmlHierarchyConfigurator.cs
@@ -19,7 +19,9 @@
 
 using System;
 using System.Collections;
+using System.Collections.Generic;
 using System.Globalization;
+using System.Linq;
 using System.Reflection;
 using System.Xml;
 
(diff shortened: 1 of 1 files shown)


dpsenner:
This is actually a great idea. With a few test cases this could make a great improvement.

jbaehr:
while I like the general idea, I have concerns regarding the ctor resolution. The current proposal works only with ctor arguments of type `String` and uses their names to find the right ctor. C#, however, uses the arugment's type for overload resolution. In addition, how does this approach mix with the existing, "by property" initialization?
I'd prefer a syntax compatible with the existing [xml parametritazion](http://logging.apache.org/log4net/release/manual/configuration.html#Parameters) and following the same rules as C# for overload resolution. What about this:

currnetly
```
<param name="evaluator" type="log4net.spi.LevelEvaluator">
    <param name="Threshold" value="WARN"/>
<param>
```
and
```
<evaluator type="log4net.spi.LevelEvaluator">
    <threshold value="WARN"/>
<evaluator>
```
both translate to the following C# equvalent:
```
evaluator = new log4net.spi.LevelEvaluator { Threshold = "WARN" };
```

To achieve something like this
```
foo = new my.Foo(arg1: 22) { Bar = "baz" };
```
we could have this extension to the XML parametrization:
```
<param name="foo" type="my.Foo">
    <constructor>
        <param type="System.Int32" name="arg1" value="22"/><!-- name is not required, but may help the human reading the xml -->
    <constructor>
    <param name="Bar " value="baz"/>
<param>
```
or, using the compact parameter syntax
```
<foo type="my.Foo">
    <constructor>
        <arg1 type="System.Int32" value="22"/>
    <constructor>
    <bar>baz</bar>
<param>
```

This would introduce a new child element for "param", namely "constructor", which is not 100% backward compatible. But neither is the original proposal. And a workaround in the form of `<param name="Constructor", ... />` is possible.

An implementation could collect and convert all params from the "constructor" element and feed them as array to System.Activator (there is an overload of `CreateInstance` taking the ctor arguments as `Object[]`). This one will then select the right ctor according to c#'s overload resolution rules. No need to iterate over the ctor meta data manually.

PR #4, opened 2017-06-08
Only look at relevant log files for backup size at initialization time.

dpsenner:


dpsenner:
This patch can be merged into master. It lacks new test cases and therefore we should do so iff it is a stable fix and does not break existing tests.

PR #5, opened 2017-06-08
Removed the event properties in CreateCompositeProperties

dpsenner:
In order to increase the speed up operations in almost x10 times of using properties in the output when UserName and Identity information are not needed.


dpsenner:
There is an issue for this too (LOG4NET-429).

dpsenner:
The patches available as attachments on issue LOG4NET-429 look more sensible to me. This patch simply removes this feature and thus breaks current behavior. In my eyes this is not a great deal.

PR #6, opened 2017-06-22
Try and trap a date when looking for the next backup number.

dpsenner:
This reopens the old pull request 29 and provides a place for discussion. Please note that the old pull request 43 provides the same functionality.


dpsenner:
This patch provides the same functionality as #9 does. Since this patch affects the current behavior and I have a bad feeling about changing the behavior of the rolling file appender, I see a unit test as a requirement for this pull request to proceed.

PR #7, opened 2017-06-22
Add log4net asp.net extension

dpsenner:
This reopens the old pull request 32 and provides a place for discussion. It would be nice to have some feedback from the author of the original patch.


dpsenner:
We need an ICLA from the author Peter Jas since this is a rather large change.

saeidehv:
Hi,
Will this PR be available in the next release of log4net? Is there any estimation for release date?
Thanks,

jvz:
Would it be worth creating a software grant as well if an ICLA doesn't already exist?

dpsenner:
For the readers not knowing about what a software grant is, follow [this link](http://www.apache.org/licenses/proposed/software-grant.txt).

A software grant may be desirable, given that this is a complete and working application. But it entirely depends on what the original author of the contribution likes to do.

PR #8, opened 2017-06-22
Implemented changes for the Util/AppenderAttachedImpl.cs class for calling log4net appenders in parallel using .NET Task Parallel Library

dpsenner:
This reopens the old pull request 40 and provides a place for discussion.


dpsenner:
This is again a rather large patch that requires an ICLA from the original author Harry Martyrossian.

PR #9, opened 2017-06-22
Allow for dates that aren't necessarily in the yyyy-MM-dd format.

dpsenner:
This reopens the old pull request 43. Please note that the old pull request 29 (reopened as #6) provides the same functionality.


dpsenner:
With respect to #6 this patch is cleaner. Since this affects the current behavior and I have a bad feeling about changing the behavior of the rolling file appender, I see a unit test as a requirement for this pull request to proceed.

PR #10, opened 2017-06-22
Improve the performance of looking up the windows identity

dpsenner:
This reopens the old pull request 45.


dpsenner:
The patch basically wraps the invocation to WindowsIdentity.GetCurrent() into a provider that refreshes the identity in the background once every 15 seconds. Please discuss!

PR #11, opened 2017-06-29
(doc) Fix name of class LogicalThreadContext

MarcelGosselin:
Fix name of class LogicalThreadContext in Scopes section, words were reversed.


bodewig:
merged, many thanks!

PR #15, opened 2017-08-18
https://issues.apache.org/jira/browse/LOG4NET-552

ssachin3108:
services installed on production box are separate process and are configured to write log into single log file. Due to this following two issues are happening,

During write/append operation some processes are failing with error “Unable to acquire lock on file. The process cannot access the file because it is being used by another process”. This is because one process acquires lock (thread safe not process safe) for writing into log file and simultaneously another is trying to acquire lock and fails with above error and it just skips writing into log file.

During rolling operation, rolled file gets created with less than 1KB size. Thus log entries are lost. Upon investigation we found that, rolling operation is protected by system wide Mutex lock. At the time of rolling, multiple processes may come at the same time for rolling and first process will roll the original file correctly and give a different name to rolled file and re-create blank original file. Now the second process which would have come for rolling will roll the blank original file and overwrite the rolled file created by first process and thus rolled file is losing the data.

We locally have fixed above issues by changing latest log4net source code. We have kept locking model as “MinimalLock” which is current configuration in production. During append operation for acquirelock/releaselock we added system wide mutex so that each process will wait for other process to complete append operation. Thus it will not skip log from being written.

During rolling operation we added check whether rolling is already happened by some other process. If yes then skip rolling operation. This will resolve rolling file overwrite by other process issue.

### DIFF
--- a/src/Appender/FileAppender.cs
+++ b/src/Appender/FileAppender.cs
@@ -620,22 +620,23 @@ namespace log4net.Appender
 			private string m_filename;
 			private bool m_append;
 			private Stream m_stream = null;
-
-			/// <summary>
-			/// Prepares to open the file when the first message is logged.
-			/// </summary>
-			/// <param name="filename">The filename to use</param>
-			/// <param name="append">Whether to append to the file, or overwrite</param>
-			/// <param name="encoding">The encoding to use</param>
-			/// <remarks>
-			/// <para>
-			/// Open the file specified and prepare for logging.
-			/// No writes will be made until <see cref="AcquireLock"/> is called.
-			/// Must be called before any calls to <see cref="AcquireLock"/>,
-			/// <see cref="ReleaseLock"/> and <see cref="CloseFile"/>.
-			/// </para>
-			/// </remarks>
-			public override void OpenFile(string filename, bool append, Encoding encoding)
+            private Mutex m_appendMutex = null;
+            private string m_appendMutexFriendlyName = "MUTEX_INTERPROCESS_COMMUNICATION_ACROSS_PROCESSES";
+            /// <summary>
+            /// Prepares to open the file when the first message is logged.
+            /// </summary>
+            /// <param name="filename">The filename to use</param>
+            /// <param name="append">Whether to append to the file, or overwrite</param>
+            /// <param name="encoding">The encoding to use</param>
+            /// <remarks>
+            /// <para>
+            /// Open the file specified and prepare for logging.
+            /// No writes will be made until <see cref="AcquireLock"/> is called.
+            /// Must be called before any calls to <see cref="AcquireLock"/>,
+            /// <see cref="ReleaseLock"/> and <see cref="CloseFile"/>.
+            /// </para>
+            /// </remarks>
+            public override void OpenFile(string filename, bool append, Encoding encoding)
 			{
 				m_filename = filename;
 				m_append = append;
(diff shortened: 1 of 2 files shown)


src/Appender/FileAppender.cs
```diff
@@ -671,7 +672,15 @@ public override Stream AcquireLock()
 				{
 					try
 					{
-						m_stream = CreateStream(m_filename, m_append, FileShare.Read);
+                        if (m_appendMutex == null)
+                        {
+                            if (!Mutex.TryOpenExisting(m_appendMutexFriendlyName, out m_appendMutex))
+                                m_appendMutex = new Mutex(false, m_appendMutexFriendlyName);
```
dpsenner: Please wrap this in braces (`{}`)

src/Appender/FileAppender.cs
```diff
@@ -695,7 +704,10 @@ public override void ReleaseLock()
 			{
 				CloseStream(m_stream);
 				m_stream = null;
-			}
+
+                if (m_appendMutex != null)
+                    m_appendMutex.ReleaseMutex();
```
dpsenner: Please wrap this in braces `{` and `}`.

src/Appender/FileAppender.cs
```diff
@@ -620,22 +620,23 @@ public class MinimalLock : LockingModelBase
...
-			/// <para>
-			/// Open the file specified and prepare for logging.
-			/// No writes will be made until <see cref="AcquireLock"/> is called.
-			/// Must be called before any calls to <see cref="AcquireLock"/>,
-			/// <see cref="ReleaseLock"/> and <see cref="CloseFile"/>.
-			/// </para>
-			/// </remarks>
-			public override void OpenFile(string filename, bool append, Encoding encoding)
+            private Mutex m_appendMutex = null;
+            private string m_appendMutexFriendlyName = "MUTEX_INTERPROCESS_COMMUNICATION_ACROSS_PROCESSES";
```
dpsenner: Should this be a more random string? `MUTEX_INTERPROCESS_COMMUNICATION_ACROSS_PROCESSES` might be in use by anyone. The mutex name should at least contain `log4net` and a scope what it is used for. Interprocess communication is not what the mutex locks and thus not the correct scope.

dpsenner: Should this be a more random string? `MUTEX_INTERPROCESS_COMMUNICATION_ACROSS_PROCESSES` might be in use by anyone. The mutex name should at least contain `log4net` and a scope what it is used for. Interprocess communication is not what the mutex locks and thus not the correct scope.

src/Appender/FileAppender.cs
dpsenner: Please wrap this in braces (`{` and `}`)

src/Appender/FileAppender.cs
dpsenner: Please wrap this in braces (`{` and `}`).

src/Appender/FileAppender.cs
```diff
@@ -620,22 +620,23 @@ public class MinimalLock : LockingModelBase
...
+            /// <param name="filename">The filename to use</param>
+            /// <param name="append">Whether to append to the file, or overwrite</param>
+            /// <param name="encoding">The encoding to use</param>
+            /// <remarks>
+            /// <para>
+            /// Open the file specified and prepare for logging.
+            /// No writes will be made until <see cref="AcquireLock"/> is called.
+            /// Must be called before any calls to <see cref="AcquireLock"/>,
+            /// <see cref="ReleaseLock"/> and <see cref="CloseFile"/>.
+            /// </para>
```
dpsenner: Update the docstring to document the new behavior.

PR #16, opened 2017-08-19
Implemented review comments given under Pull Request

ssachin3108:
Implemented review comments given under Pull Request 
https://github.com/apache/logging-log4net/pull/15

Below was question asked in above PR,

Question: Are you aware of that a mutex lock when opening a file has a significant performance impact when using minimal lock scenario? 
ANS: Yes we are aware about this. But as our customer want to have multiple processes writing into same file using RollingFileAppender, we choose to implement mutex in minimal lock model. But we have make usage of this mutex configurable.

Question: Would the patch also work without the mutex but only the fix in the RollingFileAppender class?
Ans: Yes we have added new configuration 
By Default mutex won't be applied with minimal lock unless above config key is set to true.

### DIFF
--- a/src/Appender/FileAppender.cs
+++ b/src/Appender/FileAppender.cs
@@ -620,22 +620,23 @@ namespace log4net.Appender
 			private string m_filename;
 			private bool m_append;
 			private Stream m_stream = null;
-
-			/// <summary>
-			/// Prepares to open the file when the first message is logged.
-			/// </summary>
-			/// <param name="filename">The filename to use</param>
-			/// <param name="append">Whether to append to the file, or overwrite</param>
-			/// <param name="encoding">The encoding to use</param>
-			/// <remarks>
-			/// <para>
-			/// Open the file specified and prepare for logging.
-			/// No writes will be made until <see cref="AcquireLock"/> is called.
-			/// Must be called before any calls to <see cref="AcquireLock"/>,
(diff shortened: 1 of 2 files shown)


src/Appender/FileAppender.cs
```diff
@@ -371,21 +371,22 @@ public abstract class LockingModelBase
...
+            /// <param name="useMutexWithMinimalLock">Whether to protect append operation in Minimal lock with Mutex </param>
+            /// <remarks>
+            /// <para>
+            /// Open the file specified and prepare for logging.
+            /// No writes will be made until <see cref="AcquireLock"/> is called.
+            /// Must be called before any calls to <see cref="AcquireLock"/>,
+            /// <see cref="ReleaseLock"/> and <see cref="CloseFile"/>.
+            /// </para>
+            /// </remarks>
+            public abstract void OpenFile(string filename, bool append, Encoding encoding, bool useMutexWithMinimalLock = false);
```
jbaehr: Please do not change the public API of the locking model. This would make all locking model implementations out there incompatible with the next log4net release. Prefer a new locking model instead.

A basic idea of the strategy pattern for the locking model is that the class using the strategy (`FileAppender`, in our case) does not require any knowledge about a specific strategy. With the new setting `MutexWithMinimalLock` however, such a locking model specific knowledge is introduced. And it requires changes to all other implementations as well.

PR #17, opened 2017-11-09
Save one frame from being accessed twice unnecessarily in ctor of LocationInfo

shubagarwal:
The last frame in stack with declaringType as callerStackBoundaryDeclaringType was being accessed in both while loops.


dpsenner:
Would you please explain the changeset more verbosely?

shubagarwal:
Update the description in first comment. Please check.

PR #18, opened 2018-04-16
Update XmlLayoutSchemaLog4j.cs

dmarlow:
Use prefix and ns to play nice with XmlWriter

### DIFF
--- a/src/Layout/XmlLayoutSchemaLog4j.cs
+++ b/src/Layout/XmlLayoutSchemaLog4j.cs
@@ -172,7 +172,7 @@ method="run" file="Generator.java" line="94"/>
 			}
 
 			// Write the start element
-			writer.WriteStartElement("log4j:event");
+			writer.WriteStartElement("log4j", "event", "log4net");
 			writer.WriteAttributeString("logger", loggingEvent.LoggerName);
 
 			// Calculate the timestamp as the number of milliseconds since january 1970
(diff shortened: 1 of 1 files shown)


dmarlow:
I re-made the edits as the previous commit had white-space issues. This one is much friendlier.

Vizehase:
We support this pull request. Without this bugfix, the layout does not work on Win .NET Core 3.1 platform (nothing is actually logged any more), so we had to take this bugfix into an own layout class and redirect our config files to that. Please merge this pull request!

StarWars999123:
I highly doubt, that this PR should have been closed. I currently assume, that the RenderLoggingEvent method might have challenges to keep up with this change. 
It could be, that the additional log4net element can not be parsed.

tobylo:
I'm going to assume this change is the reason we no longer get XML logs for our .NET framework services after upgrading from 2.0.12 to 2.0.14. Works for netcoreapp, net5 and net6.

net48 and net461 (the ones I've tried with) results in blank xml files.

fluffynuts:
@tobylo I've been looking into this as part of another issue - there's a similar situation in the "non-log4j" variant which I had to update for the test app I wrote to work against it since the test app is using dotnet 6. I have updates in the pipeline that use the merged behavior for the NETSTANDARD target only, reverting to the original behavior otherwise, which should solve the problem for both use-cases.

This is an unfortunate case of dotnet breaking compatibility - older `XmlWriter.WriteStartElement` calls would allow element names with a `:` in them (ie `log4j:event`) where dotnet core / 5+ throw an argument exception for this and provide no way to request the original behavior - instead,  you have to go all-in on namespacing. I don't know when I'll get a release out because I was working on this to solve another issue (issue claims that there's a circumstance where logs are overwritten instead of rolled, and the example provides uses the other xml layout class). Perhaps it's worthwhile doing a fix release _just_ to get this behavior out? Your thoughts?

Aaron-Ritter-IDEXX:
I agree that it would have been helpful to see a mention in the release notes, here. I took an educated guess to click on the link to this PR, and luckily, I was right. It was a good thing we have some unit/integration tests for our logging, as weird as that sounds. https://logging.apache.org/log4net/release/release-notes.html

PR #21, opened 2018-05-09
Wip async forwarder1

wjdavis5:
Example 1 of a potential AsyncForwardingAppender

This example adds:
```
internal Task<int> AppendLoopOnAppendersAsync(LoggingEvent[] loggingEvents)
public Task<int> AppendLoopOnAppendersAsync(LoggingEvent loggingEvent)
```


wjdavis5:
Well yes I see what you're saying. I'll get a 2nd example up soon that you'll hopefully consider as an interim.
Long term I could see creating an `IAsyncAppender` interface that implements `IAppender`.

PR #22, opened 2018-05-30
LOG4NET-586 provide log4j like XML Layout that works on .NET Core 1.x

bodewig:


dpsenner:
There is a lot of whitespace changes in this PR which makes it tough to review. Would you mind to apply those changes onto develop and only keep the changes/features in this PR?

src/Layout/XmlLayoutBaseNS.cs
```diff
@@ -0,0 +1,281 @@
...
+	/// <para>
+	/// This is an abstract class that must be subclassed by an implementation
+	/// to conform to a specific schema.
+	/// </para>
+	/// <para>
+	/// Deriving classes must implement the <see cref="FormatXml"/> method.
+	/// </para>
+	/// </remarks>
+	abstract public class XmlLayoutBaseNS : LayoutSkeleton
+	{
```
dpsenner: Would you please explain the reasoning behind this base class? As far as I can tell we could reduce both the inheritance complexity and the public api without it. I discourage inheritance and encourage composition when there is need for it.

src/Layout/XmlLayoutSchemaLog4jNS.cs
```diff
@@ -0,0 +1,282 @@
...
+	/// </summary>
+	/// <remarks>
+	/// <para>
+	/// Formats the log events according to the
+	/// http://logging.apache.org/log4j schema and actually puts the
+	/// elements into a namespace. This may break tooling that doesn't
+	/// handle XML using namespaces.
+	/// </para>
+	/// </remarks>
+	/// <author>Nicko Cadell</author>
```
dpsenner: You are not nicko, are you? 😋

src/Layout/XmlLayoutSchemaLog4j.cs
```diff
@@ -28,7 +28,7 @@
 namespace log4net.Layout
 {
 	/// <summary>
-	/// Layout that formats the log events as XML elements compatible with the log4j schema
+	/// Layout that formats the log events as XML elements compatible with the log4j 1.2 schema
```
dpsenner: The newly added layout class claims to do this, too. Would you please explain this change?

src/Layout/XmlLayoutSchemaLog4jNS.cs
```diff
@@ -0,0 +1,282 @@
...
+	public class XmlLayoutSchemaLog4jNS : XmlLayoutBaseNS
```
dpsenner: Would we like to have the version in the class name, too? If tomorrow there is to be 1.3 layout we have no place to put that.

bodewig:
The whitespace changes are inside a separate commit - ec82f5b79b0c2d09122fa9cc7bf8281dc9ac93e5 - I'll cherry-pick it to develop and rebase this branch.

src/Layout/XmlLayoutBaseNS.cs
bodewig: Initially I planned to also create a replacement of `XmlLayout` as it suffers from the same problems if you specify a prefix. I'm not sure this is something people need anymore, in which case we can merge the classes.

src/Layout/XmlLayoutBaseNS.cs
```diff
@@ -0,0 +1,281 @@
...
+		public string InvalidCharReplacement
+		{
+			get {return m_invalidCharReplacement;}
+			set {m_invalidCharReplacement=value;}
+		}
+
+		/// <summary>
+		/// The namespace URI to use for the elements and attributes written by this layout.
+		/// </summary>
+		public virtual string NamespaceUri
```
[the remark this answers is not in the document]
bodewig: I've decided to override the getters rather than setting default values in `ActivateOptions` or the constructor.

src/Layout/XmlLayoutSchemaLog4j.cs
bodewig: the class has a version field for historical reasons but only accepts 1.2 as value. To me it is more important to say it is the 1.x version and not the 2.x version of log4j.

src/Layout/XmlLayoutSchemaLog4jNS.cs
bodewig: but he's written most of the code I copied :-)

src/Layout/XmlLayoutSchemaLog4jNS.cs
bodewig: 2.0 rather than 1.3 - I wanted to keep symmetry with the existing class name.

src/Layout/XmlLayoutSchemaLog4jNS.cs
dpsenner: That's probably fine, then.

src/Layout/XmlLayoutBaseNS.cs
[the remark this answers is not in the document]
dpsenner: Wouldn't that make any customization of this class' behavior impossible? Should the class be marked sealed?

src/Layout/XmlLayoutBaseNS.cs
dpsenner: Does this indicate that we should rather implement a `XmlFormatter` class that can be plugged into various appenders? This is the composition what I have in mind and that we should prefer over inheritance.

src/Layout/XmlLayoutBaseNS.cs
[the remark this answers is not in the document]
bodewig: it's about providing a default value, not really customizing anything. I'd move it to the constructor and make the properties non-virtual unless you prefer to get rid of the base class completely.

src/Layout/XmlLayoutBaseNS.cs
bodewig: The layout is what gets plugged into the appenders, isn't it? The whole structure of `XmlLayoutBaseNS` and `XmlLayoutSchemaLog4jNS` just mimics what has been there before - an area of code I never really looked into so far.

src/Layout/XmlLayoutSchemaLog4jNS.cs
bodewig: The responsibility of the layout is formatting, I don't really see what an additional formatter would do. Please note that we need to configure `XmlWriter` in different ways for the "no namespace" and "with namespace" cases.

src/Layout/XmlLayoutBaseNS.cs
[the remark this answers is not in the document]
bodewig: changed by now.

PR #27, opened 2018-10-01
 Added tests for the RollingLockStrategy

astryia:


dpsenner:
Awesome, I'm going to look at it later today.

tests/src/Appender/RollingFileAppenderTest.cs
```diff
@@ -1747,17 +1773,93 @@ public void TestDefaultLockingModel()
 			DestroyLogger();
 		}
 
-		[Test, Ignore("Not Implemented: this test should assert that the rolling file appender works when configured with rolling lock strategy none")]
-		public void TestRollingLockStrategyNone()
+		/// <summary>
+		/// Verifies that the rolling file appender works when configured with rolling lock strategy none and local mutex
```
dpsenner: A bit nitpicky, but " and local mutex" should be scratched from the summary.

astryia: I can update summary as following
**"Verifies that the rolling file appender works when configured with rolling lock strategy"**
since there are two test cases there for _RollingLockStrategyKind.LocalMutex_ and for _RollingLockStrategyKind.None_

dpsenner: That's great, thanks! I am nitpicky because this information is going to stay for long. It is therefore important to reflect what the test is about, avoiding confusion in future interpretations.

astryia: Done

PR #29, opened 2018-11-22
Parameterless LogManager.GetLogger

hanabi1224:
A parameterless LogManager.GetLogger() method to simplify common usage:

private static readonly ILog logger = LogManager.GetLogger(System.Reflection.MethodBase.GetCurrentMethod().DeclaringType);

->

private static readonly ILog logger = LogManager.GetLogger();

### DIFF
--- a/src/LogManager.cs
+++ b/src/LogManager.cs
@@ -18,6 +18,7 @@
 #endregion
 
 using System;
+using System.Diagnostics;
 using System.Reflection;
 
 using log4net.Core;
@@ -259,6 +260,20 @@ namespace log4net
 #else
 			return GetLogger(Assembly.GetCallingAssembly(), type.FullName);
 #endif
+		}
+
+		/// <summary>
+		/// Shorthand for <see cref="M:LogManager.GetLogger()"/>.
+		/// </summary>
+		/// <remarks>
+		/// Get the logger for the fully qualified name of the type of the caller.
+		/// LogManager.GetLogger() is equivalent to LogManager.GetLogger(System.Reflection.MethodBase.GetCurrentMethod().DeclaringType);
+		/// </remarks>
+		/// <returns>The logger with the name specified.</returns>
+		public static ILog GetLogger()
+		{
+			var callingMethod = new StackTrace().GetFrame(1).GetMethod();
+			return GetLogger(callingMethod.DeclaringType);
 		}
 
 		/// <summary>


dpsenner:
Hi hanabi,

thanks for your contribution. While your short form looks handy, it also adds a public api that relies on examining the call stack to guess a logger name. People could argue that they would like to guess the logger name from the next outer scope. Therefore I wonder whether another solution could satisfy your needs. For instance, I usually write this such as:

```csharp
private static readonly ILog logger = LogManager.GetLogger(typeof(ContainingType));
```

and further automate that line of code with a template in visual studio.

Best regards,
Dominik

hanabi1224:
@dpsenner Thank you for your reply, that is good if it comes with a CA rule to warn if people copy that line across classes but forget to replace the type,  because ideally every class has this static logger instance, it's common that people do the copy paste (especially during refactoring), and when these classes share namespaces, compiler won't even complain.   Please let's me know your concern.

dpsenner:
Essentially, what you are looking for is a static method that gets a logger based on your project's / company's guidelines. The following would get the job done without extension of the logging framework's api:

```csharp
public static class MyCompanyLoggerPolicy {
    public static ILogger GetLogger() {
        var callingMethod = new StackTrace().GetFrame(1).GetMethod();
        return GetLogger(callingMethod.DeclaringType);
    }
}
```

And the usage is rather clear:

```csharp
public class MyClass {
   private static readonly ILogger Logger = MyCompanyLoggerPolicy.GetLogger();
}
```

While I appreciate the effort you are putting into this contribution, I do not think that this guideline applies to the major part of the users and therefore I am currently -0 for this contribution. Please do start a thread on the mailing lists (dev@logging.apache.org and log4net-user@logging.apache.org and general@logging.apache.org) to start a discussion about this change. This way you should reach a large audience and we will eventually know what the rest of the community thinks about this contribution.

hanabi1224:
@dpsenner  thanks, will do

jziolkow:
On this subject, the following method is no longer available in v.2.0.8:

```
public static ILog GetLogger(string name)
		{
			return GetLogger(Assembly.GetCallingAssembly(), name);
		}
```

Looking at the source, it looks like it's being excluded by a check for .NET Standard 1.3. I'm trying to use the library with .NET Core 2.1/.NET Standard 2.0. Can the method be restored? Without the method, code that I don't have access to modify is failing with a MissingMethodException.

dpsenner:
log4net still targets netstandard-1.3 and not netstandard-2.0, therefore the answer is no. A workaround that works for netstandard-1.3 would be a nice contribution.

PR #40, opened 2018-12-07
This fixes the Mutex name in the RollingFileAppender

Shazwazza:
This fixes the Mutex name in the RollingFileAppender and the FileAppender to use the ApplicationUniqueId since the base file path isn't enough in hosted web applications

There is a similar/same bug already logged for this issue here https://issues.apache.org/jira/browse/LOG4NET-506

This fix will specifically fix the issue where the rolling file appender will not log for any websites running on the same server with the same log file path and for sites on Azure this can be common, for example we've had the problem where several of our websites weren't logging because when on the same server they all have the same log file path (i.e. c:\home\site\wwwroot\app_data\log.txt)

This fix won't fix the issue where the same base path applies to applications on the same server that are not websites since it's using the AppDomainAppId to differentiate.


dpsenner:
Hi Shazwazza, thanks for your valuable contribution. Unfortunately it does not pass the continuous integration. Please rebase your changes onto develop. Ideally you also add a test to the contribution that hardens the addition against regressions.

Shazwazza:
Hi, sorry for the delay, i still have this starred in my inbox and will get back to it soon hopefully

Shazwazza:
Hi @dpsenner , I've pushed some updates

* Rebased on the develop branch and it's up to date
* Centralized mutex lock naming code
* Added a test for generating the mutex lock name
* I had to update the test project to the latest Nunit packges so i could run the unit tests in Visual Studio

Shazwazza:
Any updates on this?

rgoers:
Please see the latest emails on the logging developer's list.

PR #41, opened 2018-12-11

pctimhk:
I cannot perform the test in my visual studio 2013 environment. The feature is not yet test.


dpsenner:
Please rebase your changes onto develop. Jenkins will pick up the task and build your contribution. It will also run the tests.

PR #42, opened 2018-12-12

pctimhk:
I cannot perform the test in my visual studio 2013 environment. The feature is not yet test.


dpsenner:
The contribution does not build against the netstandard-1.3 target.

dpsenner:
There is a newer pull request with the same title. Can this pull request be closed because it is superseded by #43?

PR #43, opened 2018-12-13

pctimhk:
Enhance the filter to support netstandard-1.3.

### DIFF
--- a/src/Filter/ExceptionTypeFilter.cs
+++ b/src/Filter/ExceptionTypeFilter.cs
@@ -1,0 +1,54 @@
+﻿using log4net.Core;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace log4net.Filter
+{
+    public class ExceptionTypeFilter : FilterSkeleton
+    {
+        public ExceptionTypeFilter()
+        {
+            this.AcceptOnMatch = true;
+        }
+
+        /// <summary>
+        /// Set the exception type name to filter
+        /// </summary>
+        public string ExceptionTypeName { get; set; }
+
+        /// <summary>
+        /// <see cref="FilterDecision.Accept"/> when matching <see cref="ExceptionTypeName"/>
+        /// Default value is true.
+        /// </summary>
+        public bool AcceptOnMatch { get; set; }
+
+        public override FilterDecision Decide(LoggingEvent loggingEvent)
+        {
+            var myExceptionType = Type.GetType(this.ExceptionTypeName, false);
+
+            if (this.ExceptionTypeName != null
+                && myExceptionType != null
+                && loggingEvent.ExceptionObject != null)
+            {
+
+                var myIsMatched = myExceptionType.IsInstanceOfType(loggingEvent.ExceptionObject);
+
+                if (this.AcceptOnMatch)
+                {
+                    if (myIsMatched)
+                        return FilterDecision.Accept;
+                }
+                else
+                {
+                    if (myIsMatched)
+                        return FilterDecision.Deny;
+                }
+            }
+
+            return FilterDecision.Neutral;
+        }
+    }
+}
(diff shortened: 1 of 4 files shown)


src/Filter/ExceptionTypeFilter.cs
```diff
@@ -0,0 +1,54 @@
...
+        /// </summary>
+        public string ExceptionTypeName { get; set; }
+
+        /// <summary>
+        /// <see cref="FilterDecision.Accept"/> when matching <see cref="ExceptionTypeName"/>
+        /// Default value is true.
+        /// </summary>
+        public bool AcceptOnMatch { get; set; }
+
+        public override FilterDecision Decide(LoggingEvent loggingEvent)
```
dpsenner: Please add documention that explains the expected behavior. Any public api endpoint requires a docstring.

src/Filter/ExceptionTypeFilter.cs
```diff
@@ -0,0 +1,54 @@
...
+    {
+        public ExceptionTypeFilter()
+        {
+            this.AcceptOnMatch = true;
+        }
+
+        /// <summary>
+        /// Set the exception type name to filter
+        /// </summary>
+        public string ExceptionTypeName { get; set; }
```
dpsenner: These properties are not supported on ancient frameworks like .net 2.0. By changing the getter and setter to rely on a backing field this contribution will also build against .net 2.0. Further it allows the implementation to resolve the type behind the string and do the expensive reflection only once and not for any future log event.

src/Filter/ExceptionTypeFilter.cs
```diff
@@ -0,0 +1,54 @@
...
+        public override FilterDecision Decide(LoggingEvent loggingEvent)
+        {
+            var myExceptionType = Type.GetType(this.ExceptionTypeName, false);
+
+            if (this.ExceptionTypeName != null
+                && myExceptionType != null
+                && loggingEvent.ExceptionObject != null)
+            {
+
+                var myIsMatched = myExceptionType.IsInstanceOfType(loggingEvent.ExceptionObject);
```
dpsenner: I prefer this local variable to be refactored to read `isMatch`.

src/Filter/ExceptionTypeFilter.cs
```diff
@@ -0,0 +1,54 @@
...
+
+                if (this.AcceptOnMatch)
```
dpsenner: Swapping the comparison logic greatly improves the readability of what this filter actually does:

```csharp
if (isMatch)
{
   if (AcceptOnMatch)
   {
      return FilterDecision.Accept;
   }
   else
   {
      return FilterDecision.Deny;
   }
}
```

src/Filter/ExceptionTypeFilter.cs
```diff
@@ -0,0 +1,54 @@
...
+            this.AcceptOnMatch = true;
+        }
+
+        /// <summary>
+        /// Set the exception type name to filter
+        /// </summary>
+        public string ExceptionTypeName { get; set; }
+
+        /// <summary>
+        /// <see cref="FilterDecision.Accept"/> when matching <see cref="ExceptionTypeName"/>
```
dpsenner: To me the docstring is inconclusive on what this property actually does. It should contain at least a description of what happens when this value is set to "true" and what happens when this value is set to "false". The reasoning is that the documentation is auto-generated to be available online and hence the source of truth for any future reader that would like to use this filter. Future readers should not be required to look at the implementation.

tests/src/Filter/ExceptionTypeFilterTest.cs
```diff
@@ -0,0 +1,118 @@
...
+            </log4net>");
+            #endregion
+
+            ILoggerRepository rep = LogManager.CreateRepository(Guid.NewGuid().ToString());
+            XmlConfigurator.Configure(rep, log4netConfig["log4net"]);
+
+            loggerRepository = LogManager.CreateRepository(Guid.NewGuid().ToString());
+        }
+
+        [Test]
```
dpsenner: The test only covers the default path of "AcceptOnMatch", which happens to be "AcceptOnMatch=true". It should also cover the paths where a user sets AcceptOnMatch to be true or false explicitly. With this in place, future changes to that behavior are detected as regressions.

PR #44, opened 2019-01-17
Performance fix for lookup up windows username

davebfid:
This should fix LOG4NET-429 and I think it is a better approach than everything else which has been suggested.

The approach is when the user name is requested, to cache the username and the sid of the current user, the next time user name is requested if the sid hasn't changed then returned the cached user name

Internally in .NET when you call WindowsIdentity.GetCurrent().Name it takes the sid and translates it into the account name, so this approach should work fine.


dpsenner:
Please rebase your changes onto develop.

davebfid:
I'll submit a new pull request based on develop later today

PR #45, opened 2019-01-22
Fix performance issue LOG4NET-429 that was introduced by LOG4NET-205

davebfid:
This pull request fixes the performance issue raised as LOG4NET-429.  I think it is a better solution than the potential fixes mentioned in LOG4NET-429.  Also I feel its better solution than the one proposed in pull request 10.

The change caches 3 things:
Name
SID
When the name was looked up

When the user name is required the code performs the following checks:
If the cached SID is the same as the current user SID:
if its not then the name is looked up again.

If cached details are older than 10 seconds:
If they are then the name is looked up again.
This ensures if a user is renamed (in which case the SID remains the same but the user name will be different) it will be no more than 10 seconds out of date

### DIFF
--- a/src/Core/LoggingEvent.cs
+++ b/src/Core/LoggingEvent.cs
@@ -852,7 +852,7 @@ namespace log4net.Core
 		{
 			get
 			{
-				if (m_data.UserName == null  && this.m_cacheUpdatable) 
+				if (m_data.UserName == null  && this.m_cacheUpdatable)
 				{
 					m_data.UserName = WindowsIdentityProvider.CurrentIdentityName;
 				}
--- a/src/Core/LoggingEvent.cs
+++ b/src/Core/LoggingEvent.cs
@@ -22,11 +22,7 @@ using System.Collections;
 using System.IO;
 #if (!NETCF)
 using System.Runtime.Serialization;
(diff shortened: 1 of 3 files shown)


dpsenner:
Thanks for rebasing. Looks like now there are some syntax glitches that fail the build.

dpsenner:
I like the exhaustive information backing up the contribution, awesome!

davebfid:
Somewhat struggling to see how a ThreadContextTest would fail for a UserName change

dpsenner:
Indeed an interesting fail. I just started a build of the develop branch to see if that fails too and another build of this pr to investigate any flakyness. This is going to take a while. Build results will be available [here](https://builds.apache.org/blue/organizations/jenkins/logging-log4net/activity) in the future.

dpsenner:
The test fails on the develop branch, too. Would you like to investigate and fix it in a separate pull request?

davebfid:
Will be a few days before I get chance but I should be able to look into it.
I'll also edit the commit message and submit some tests for my change at the same time.

msallin:
This blocks our team for years now... Could we contribute somehow?

msallin:
@davebfid can we support somehow, to bring this forward?

PR #46, opened 2019-01-29
'(doc)' make the method RollingFileAppender.RollFile() virtual.

qub1n:
I need to post process rolled file by archiver and I can hardly achieve it without making this method virtual.

### DIFF
--- a/src/Appender/RollingFileAppender.cs
+++ b/src/Appender/RollingFileAppender.cs
@@ -1269,7 +1269,7 @@ namespace log4net.Appender
 		/// also checks for existence of target file and deletes if it does.
 		/// </para>
 		/// </remarks>
-		protected void RollFile(string fromFile, string toFile) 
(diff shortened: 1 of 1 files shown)


dpsenner:
Would it be more appropriate to add an invocation of an emty but virtual PostRollFile method?

qub1n:
Yes following empty method would be sufficient.
protected virtual void OnFileRolled(string filename){}

dpsenner:
I propose to add both oldFile and newFile to the method arguments. Is there a reasoning why the pull request is not based onto develop?

qub1n:
Is there a reasoning why the pull request is not based onto develop?
-> No, it was just the default option on GitHub, I will change it.

OK, please cancel this pull request and I will create a new Pull Request on branch 'develop' with following method signature:

protected virtual void OnFileRolled(string fromFile, string toFile){}

qub1n:
I made new pull request (47) on proper branch, please close it.

PR #48, opened 2019-05-22
Fixes LOG4NET-587 Mutex ~ UnauthorizedAccessException Access to the path is denied

mirec75:
This pull request fixes [LOG4NET-587](https://issues.apache.org/jira/browse/LOG4NET-587) issue: Mutex ~ UnauthorizedAccessException Access to the path is denied. 

I fixed this issue as suggested by [Dominik Psenner](https://issues.apache.org/jira/browse/LOG4NET-587?focusedCommentId=16326387&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-16326387) see the fix suggestion [here](https://stackoverflow.com/questions/19536697/unauthorizedaccessexception-when-trying-to-open-a-mutex).

I experienced this bug often, until I fixed and successfully tested it.
We have many .Net 4.0 Web Services (WCF) and ASPX pages hosted in IIS, where I newly integrated log4net for a sake of quick high level error tracing, as a last resort if the legacy functionality does not trace anything. 
After fixing the error above and testing I simplified things and went to using the base FileAppender, because of several further issues with rolling, so the Mutexes are no more an issue also with the currently latest stable log4net 2.0.8

There seem to be several similar issues as well, which may be worth checking with this fix, e.g.: LOG4NET-496, LOG4NET-506, LOG4NET-525, LOG4NET-560, LOG4NET-599, LOG4NET-612

### DIFF
--- a/src/Appender/FileAppender.cs
+++ b/src/Appender/FileAppender.cs
@@ -888,7 +888,7 @@ namespace log4net.Appender
 				return null;
 			}
 
-			return Regex.Replace(filePath, "[" + new string(Path.GetInvalidFileNameChars()) + "]", "_");
+			return Regex.Replace(filePath, "[" + new string(Path.GetInvalidFileNameChars()).Replace(@"\", @"\\") + "]", "_");
 		}
 
 		/// <summary>
--- a/src/Appender/FileAppender.cs
+++ b/src/Appender/FileAppender.cs
@@ -24,6 +24,10 @@ using System.Threading;
 using log4net.Util;
 using log4net.Layout;
 using log4net.Core;
+using System.Security.AccessControl;
+using System.Security.Principal;
+using System.Globalization;
+using System.Diagnostics;
 #if NET_4_5 || NETSTANDARD1_3
 using System.Threading.Tasks;
 #endif
@@ -723,6 +727,7 @@ namespace log4net.Appender
 		public class InterProcessLock : LockingModelBase
 		{
 			private Mutex m_mutex = null;
+			private string m_mutexName = null;
 			private Stream m_stream = null;
 			private int m_recursiveWatch = 0;
 
@@ -846,11 +851,15 @@ namespace log4net.Appender
 							.Replace(":", "_")
 							.Replace("/", "_");
 
-					m_mutex = new Mutex(false, mutexFriendlyFilename);
+					m_mutexName = mutexFriendlyFilename;
+
+					LogLog.Debug(declaringType, $"Creating mutex for file appender. Mutex name: \"{m_mutexName}\", current file: {CurrentAppender.File} .");
+
+					m_mutex = SecureCreateMutex(mutexFriendlyFilename, CurrentAppender.ErrorHandler);
 				}
 				else
 				{
-					CurrentAppender.ErrorHandler.Error("Programming error, mutex already initialized!");
+					CurrentAppender.ErrorHandler.Error($"Programming error, mutex already initialized! Mutex name: \"{m_mutexName}\", current file: {CurrentAppender.File} .");
 				}
 			}
 
@@ -867,6 +876,7 @@ namespace log4net.Appender
 					m_mutex.Close();
 #endif
 					m_mutex = null;
+					m_mutexName = null;
 				}
 				else
 				{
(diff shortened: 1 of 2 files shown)


src/Appender/FileAppender.cs
```diff
@@ -876,6 +886,57 @@ public override void OnClose()
...
+			// Using this mehotd to avoid the issue:
+			// Mutex ~ Access to the path is denied in log4net.Appender.RollingFileAppender.ActivateOptions()
+			// https://jira.apache.org/jira/browse/LOG4NET-587
+			// Solution implemented according to:
+			// UnauthorizedAccessException when trying to open a mutex
+			// https://stackoverflow.com/questions/19536697/unauthorizedaccessexception-when-trying-to-open-a-mutex
+			//NOTE: there are several other jira issues reporting this problem.
+			if (string.IsNullOrWhiteSpace(mutexId))
+			{
+				return null;
```
dpsenner: Would you please add an invocation to errorHandler such that even this rare case appears in the internal logs of the framework?

src/Appender/RollingFileAppender.cs
```diff
@@ -1695,6 +1708,8 @@ protected DateTime NextCheckDate(DateTime currentDateTime, RollPoint rollPoint)
 		/// A mutex that is used to lock rolling of files.
 		/// </summary>
 		private Mutex m_mutexForRolling;
+
+		private string m_mutexNameForRolling = null;
```
dpsenner: Storing this as an internal state appears to be unnecessary, isn't it?

src/Appender/FileAppender.cs
```diff
@@ -846,11 +851,15 @@ public override void ActivateOptions()
 							.Replace(":", "_")
 							.Replace("/", "_");
 
-					m_mutex = new Mutex(false, mutexFriendlyFilename);
+					m_mutexName = mutexFriendlyFilename;
```
dpsenner: Storing this as an internal state appears to be unnecessary, isn't it?

src/Appender/RollingFileAppender.cs
```diff
@@ -1147,8 +1148,20 @@ override public void ActivateOptions()
 			}
 
 #if !NETCF
-			// initialize the mutex that is used to lock rolling
-			m_mutexForRolling = new Mutex(false, m_baseFileName.Replace("\\", "_").Replace(":", "_").Replace("/", "_"));
+			if (m_mutexForRolling == null)
+			{
+				// initialize the mutex that is used to lock rolling
+				string mutexFriendlyFilename = m_baseFileName.Replace("\\", "_").Replace(":", "_").Replace("/", "_");
```
dpsenner: Could this dark magic be moved into the `SafeCreateMutex()`, such that it exists only once?

src/Appender/FileAppender.cs
```diff
@@ -876,6 +886,57 @@ public override void OnClose()
...
+					);
+
+				// attempt to create the mutex, with the desired DACL..
+				Mutex createdMutex = new Mutex(false, mutexId, out createdNew, mutexSecurity);
+				return createdMutex;
+			}
+			catch (WaitHandleCannotBeOpenedException ex)
+			{
+				// the mutex cannot be opened, probably because a Win32 object of a different type with the same name already exists.
+				errorHandler?.Error($"The mutex '{mutexId}' cannot be opened, probably because a Win32 object of a different type with the same name already exists.", ex);
```
dpsenner: Is this a situation where the application is still in a safe state to continue working happily? What can be done to fix the situation when this happens? It would be great if the log message included one or more hints what a user of the logging framework could do to fix this situation.

src/Appender/FileAppender.cs
```diff
@@ -876,6 +886,57 @@ public override void OnClose()
...
+			}
+			catch (UnauthorizedAccessException ex)
+			{
+				// the mutex exists, but the current process or thread token does not have permission to open the mutex with SYNCHRONIZE | MUTEX_MODIFY rights.
+				errorHandler?.Error($"The mutex '{mutexId}' exists, but the current process or thread token does not have permission to open the mutex with SYNCHRONIZE | MUTEX_MODIFY rights", ex);
```
dpsenner: What can be done to fix the situation when this happens? It would be great if the log message included one or more hints what a user of the logging framework could do to fix this situation.

src/Appender/FileAppender.cs
mirec75: yes, I'll do it.
I was thinking to skip this `if (string.IsNullOrWhiteSpace(mutexId))` altogether, or better just skip the `return null;` in this `if`, since if the `mutexId` is null, then the created mutex will be created unnamed. Yes such situation is strange i.m.h.o, and worth of reporting in log, since accessing such ambiguous mutex is error prone. 
I'll make an update to the pull request later today.

src/Appender/RollingFileAppender.cs
mirec75: I would remove it, but consider my following experience:
My reason to introduce the `m_mutexNameForRolling` was tracking the change of log file path  of the `RollingFileAppender`, after it was first correctly activated, via assigning the new file path and calling Activate() on the appender in user code again. I experienced mutex leak while recreating the mutex with a different Id, so I wanted to track this down.

Background: I know a customer doing stuff like changing the log file path in run time, due to logging to a different specific directory, based on some global DB settings, but wanting also to track other potential startup issues before the log file path was changed....

dpsenner: Such a functional requirement is better implemented by streaming the logs into a database or to a remote location. That remote location could then buffer log events in a safe manner while it is being reconfigured. The rolling file appender is, as far as I can tell, not made to handle this.

src/Appender/FileAppender.cs
dpsenner: At best this is error prone. Worst case is that rolling/logging is not synchronized and fails later on with a big bang or silently by wiping the contents of the files.

src/Appender/FileAppender.cs
mirec75: I'll remove it. See the previous comment above.

src/Appender/RollingFileAppender.cs
mirec75: ok, good to know. Thanks for clarifying this to me. I do not have much experience with streaming logs into a DB, at least not with log4net. Log4net file or console logging is for me always the "last and securest resort" to trace things in the weird cases, when you don't see any SQL or Mail messages or other App output. Log4net always delivered the essential log information to tackle errors in such situations. So, I noted, RollingFileAppender is not good to change after its first activation.

src/Appender/RollingFileAppender.cs
mirec75: in the new method, is it okay, to replace all `Path.GetInvalidFileNameChars()` not just `\/:`?

src/Appender/FileAppender.cs
mirec75: hm, this is a good question, to which I do not have an answer.
The original code did not catch the exception, thus letting the whole Appender to crash during Activate. (I think the exception was even thrown to the user code initializing log4net.) The Appender was logging nothing at all and causing all the issues like the LOG4NET-587.

If I "silently" allow to proceed (just with the invocation of the ErrorHandler), there will be another kind of issues, related with not locking. The usage of mutex itself is protected by querying of not null.

I improve the messages a bit. Give me feedback then, what do you think I shall do best.

src/Appender/FileAppender.cs
mirec75: see comment above

src/Appender/RollingFileAppender.cs
dpsenner: I am unsure whether the application still behaves well. Mutex names are not a path or a file name.

This exists to guess a mostly unique mutex name from the filename. This way another rolling file appender configured to roll another filename can happily roll at the same time.

So long this is preserved the code can do anything. Even md5 hashing would be ok.

src/Appender/FileAppender.cs
dpsenner: As far as I am concerned, applications can always be configured to log to different files and thus I would be happy with a simple file appender without locking or other fancy.

markdemich (approved):

markdemich:
Is there a reason this is not merged?  I'm having this issue.  I have multiple development app_pools in IIS that are attempting to write to the same log file and creating the same mutex.  What's odd is this used to work, but now it's not and I can't figure out what changed.  I wonder if newer versions of the .Net Framework or some windows security update broke this.  It's been a long time since I updated log4net and I'm pretty sure this was working after that so I don't think it was an update to log4net.  It seems the original mutex code was added in 2015.  We would have notice this issue way before now.

Evgentus0:
Hi @dpsenner, I am not fully understand why this pull request still is not merge? I still face with mutex issue described in this pull request. Could we merge this changes? Or it is needed some updates here?

fluffynuts:
Hi @Evgentus0 and @markdemich 
This was updated as per the changes here in the last release (2.0.13). Hope this works out well as I don't really know how to reliably test.

PR #51, opened 2019-06-13
Added support for .Net Standard 2.0

kabilanvk:
LOG4NET-572: Add target for netstandard-2.0


KamranShahid:
Needed log4net for atleast .net standard 2.0.
Problem in publishing app of .net core 3.0 where log4net 2.0,8 is referenced

kabilanvk:
> Needed log4net for atleast .net standard 2.0.
> Problem in publishing app of .net core 3.0 where log4net 2.0,8 is referenced

My change is to get a new build for .net standard 2.0 but the current build system doesn't support it. Need a new build system setup.

dpsenner:
I am afraid nant is no longer able to cope with this task. We would have to replace all what nant does with something else. This could be msbuild, paired with custom [inline build tasks written with c# code](https://docs.microsoft.com/en-us/visualstudio/msbuild/msbuild-inline-tasks?view=vs-2019). Feel free to go for it! Contributions welcome!

kabilanvk:
@dpsenner 
I don't know whether you got my emails about i've shown my interest in contributing to Log4Net. 

I have set up a sample build for Log4Net.standard solution using azure devops. Let me know if its okay to proceed then we can migrate all builds to azure devops.
https://dev.azure.com/logging-log4net/Log4Net/_build/results?buildId=20&view=artifacts

cremor:
Seems like #59 added .NET Standard 2.0 targets.

PR #53, opened 2019-10-23
LoggingEvent: repository properties get added to the composite properites

Laxo908:
Just added a few lines of code to complete this TODO. The properties of the repository are now added to the aggregated properties.


Laxo908:
It seems like Nant fails to build the commit. But I do not have any idea why.

PR #65, opened 2020-09-03
Fix: RollingFileAppender does not work on Android and iOS

triumlabs:
However netstandard contains cross-process Mutex, Android and iOS platform does not support it.
Therefore, current implementation of RollingFileAppender cannot be used on these platforms.
To fix it, Mutex object is replaced to a file based cross-process implementation.


fluffynuts:
This PR is quite stale & conflicts with the target branch; please re-open if it's still an issue you'd like to solve (: Thanks for your contribution.

PR #66, opened 2020-09-08
Fix net40 client profile

NicholasNoise:
log4net_net40cp

### DIFF
--- a/src/log4net/AssemblyInfo.cs
+++ b/src/log4net/AssemblyInfo.cs
@@ -73,14 +73,17 @@ using System.Runtime.CompilerServices;
 [assembly: AssemblyInformationalVersionAttribute("2.0.9.0-.NET 4.0")]
 [assembly: AssemblyTitle("Apache log4net for .NET Framework 4.0")]
 #endif // Client Profile
-#elif (NET_2_0)
+#elif (NET_3_5)
 #if CLIENT_PROFILE
 [assembly: AssemblyInformationalVersionAttribute("2.0.9.0-.NET 3.5 CP")]
 [assembly: AssemblyTitle("Apache log4net for .NET Framework 3.5 Client Profile")]
 #else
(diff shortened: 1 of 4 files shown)


NicholasNoise:
@fluffynuts

cremor:
Not directly related to this PR, but since you are modifying the preprocessor symbols (constants) here I thought I should mention this:
SDK style projects already define various target framework related constants automatically, see https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/preprocessor-directives/preprocessor-if#remarks

Some of those are the same as are now manually defined in the log4net.csproj and some are different. This might be confusing for developers who don't check if the constants are defined by the project itself.

But it also seems like migration to the official constants is not that easy. Because with them it seems like the version specific constants are **not** enabled for lower versions when a higher version is built.

NicholasNoise:
> @cremor 
> 
> Some of those are the same as are now manually defined in the log4net.csproj and some are different. This might be confusing for developers who don't check if the constants are defined by the project itself.
> 
> But it also seems like migration to the official constants is not that easy. Because with them it seems like the version specific constants are **not** enabled for lower versions when a higher version is built.

You're right, no other way about it. Build-in features are preferable.
But it seems unnecessary since this project is dormant.

I've spent almost two working day to enable netstandard2.0, so yes, is not easy... until old targets are supported.
Should be much easier without legacy preprocessor symbols like 'NETCF', 'CLIENT_PROFILE' and even 'NET20'

fluffynuts:
LGTM; when I have a moment, I'd like to pull down to my machine & build again to double-check; I'm taking a day off from work tomorrow, so perhaps Thursday (:

NicholasNoise:
@fluffynuts last Thursday or this one? =)

fluffynuts:
Sorry, been a bit mad -- the day off turned into a mission (:

anyway, LGTM, merging.

PR #69, opened 2020-09-29
Don't get windows username in .NET Standard 2.0

13thirteen:
Running the .NET Standard 2.0 version on Linux throws `PlatformNotSupportedException`
See https://issues.apache.org/jira/browse/LOG4NET-652

### DIFF
--- a/src/log4net/Core/LoggingEvent.cs
+++ b/src/log4net/Core/LoggingEvent.cs
@@ -895,7 +895,7 @@ namespace log4net.Core
 			{
 				if (m_data.UserName == null  && this.m_cacheUpdatable) 
 				{
-#if (NETCF || SSCLI || NETSTANDARD1_3) // NETSTANDARD1_3 TODO requires platform-specific code
+#if (NETCF || SSCLI || NETSTANDARD) // NETSTANDARD TODO requires platform-specific code
 					// On compact framework there's no notion of current Windows user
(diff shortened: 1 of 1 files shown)


src/log4net/Core/LoggingEvent.cs
```diff
@@ -895,7 +895,7 @@ public string UserName
 			{
 				if (m_data.UserName == null  && this.m_cacheUpdatable) 
 				{
-#if (NETCF || SSCLI || NETSTANDARD1_3) // NETSTANDARD1_3 TODO requires platform-specific code
+#if (NETCF || SSCLI || NETSTANDARD) // NETSTANDARD TODO requires platform-specific code
```
fluffynuts, line 898: win32 is a valid target for netstandard too -- surely there's a better alternative?

fluffynuts (changes requested):
NETSTANDARD is a valid choice for win32 targets too

src/log4net/Core/LoggingEvent.cs
fluffynuts: (for example: if the platform is !win32, we could try to invoke `whoami`)

fluffynuts:
Thanks for the contribution. This has been resolved in master and I have a release pending a vote on the dev lists for 2.0.12. I'm closing this out as people are looking for the fix. I would have preferred to have had your work & commit history so that the change would be attributed to you.

PR #70, opened 2020-10-10
Fixes LOG4NET-652 while providing non-Windows username capability

atomsmith:
This PR resolves the .NET Standard non-Windows username query issue in a different way: instead of trapping the PlatformNotSupportedException, it proactively queries whether a Windows or non-Windows platform is present. In the former case, the existing username logic is retained. In the latter case, the username is obtained through an environment query. The platform exception check that was added for .NET Standard 2.0 on non-Windows is no longer needed.


fluffynuts:
Thanks for the contribution. This has been resolved in master and I have a release pending a vote on the dev lists for 2.0.12.

PR #71, opened 2020-12-07
Fix AdoNetAppender using npgsql

NicholasNoise:
Using npgsql on netfx and mono I've encountered this error on an applicaiton startup:
```
log4net:ERROR [CustomAdoNetAppender] ErrorCode: GenericFailure. Exception while writing to database
Npgsql.PostgresException (0x80004005): 42601: syntax error at or near ":"
   at Npgsql.NpgsqlConnector.DoReadMessage(Boolean async, DataRowLoadingMode dataRowLoadingMode, Boolean readingNotifications, Boolean isPrependedMessage)
   at Npgsql.NpgsqlConnector.ReadMessage(Boolean async, DataRowLoadingMode dataRowLoadingMode, Boolean readingNotifications)
   at Npgsql.NpgsqlConnector.ReadMessage(Boolean async, DataRowLoadingMode dataRowLoadingMode, Boolean readingNotifications)
   at Npgsql.NpgsqlConnector.ReadExpecting[T](Boolean async)
   at Npgsql.NpgsqlCommand.Prepare()
   at log4net.Appender.AdoNetAppender.SendBuffer(IDbTransaction dbTran, LoggingEvent[] events)
   at log4net.Appender.AdoNetAppender.SendBuffer(LoggingEvent[] events)
```
The error prevents to write **only** very first logging event, because Npgsql fails to prepare command with no parameters (at a moment). So it is not a problem in general, but disturbing.

I think my _fix_ is not that good but does a trick.

An alternative is to disable `dbCmd.Prepare();` if it fails once, but it is overkill, cuz there is no second fail.
```
					if (m_doCommandPrepare)
					{
						try
						{
							// prepare the command, which is significantly faster
							dbCmd.Prepare();
						}
						catch (Exception)
						{
							m_doCommandPrepare = false;
						}
					}
...
		private bool m_doCommandPrepare = true;
```

### DIFF
--- a/src/log4net/Appender/AdoNetAppender.cs
+++ b/src/log4net/Appender/AdoNetAppender.cs
@@ -553,8 +553,17 @@ namespace log4net.Appender
 					{
 						dbCmd.Transaction = dbTran;
 					}
-					// prepare the command, which is significantly faster
-					dbCmd.Prepare();
+
+					try
+					{
+						// prepare the command, which is significantly faster
+						dbCmd.Prepare();
+					}
+					catch (Exception)
+					{
+						// ignore prepare exceptions as they can happen without affecting actual logging, eg on npgsql
+					}
+
 					// run for all events
 					foreach (LoggingEvent e in events)
 					{


NicholasNoise:
@fluffynuts
At a look please.

src/log4net/Appender/AdoNetAppender.cs
```diff
@@ -553,8 +553,17 @@ protected virtual void SendBuffer(IDbTransaction dbTran, LoggingEvent[] events)
...
-					dbCmd.Prepare();
+
+					try
+					{
+						// prepare the command, which is significantly faster
+						dbCmd.Prepare();
+					}
+					catch (Exception)
+					{
+						// Ignore exception
```
fluffynuts, line 564: whilst I'd like to track down the _why_ of this instead of skipping over it, I think that may take some time -- I don't even have pgsql installed on any of my machines. As far as I understand, the only side-effect here is that the command might run a little slower, so I'd say, let's put in this workaround, but please put in a comment explaining why, eg `// ignore prepare exceptions as they can happen without affecting actual logging, eg on pgsql`. Thoughts?

NicholasNoise: I've investigated it. The simple _why_ is to prepare a valid command instead of invalid one without parameters and values.
Also tracked down to inner command cache, so it is not necessary to prepare command.

fluffynuts (approved):

PR #26, opened 2018-08-10
Allow DI to create instance

pengweiqhca:

PR #72, opened 2020-12-16
Allow DI to create instances

NicholasNoise:
Replace #26

### DIFF
--- a/src/log4net/Appender/AdoNetAppender.cs
+++ b/src/log4net/Appender/AdoNetAppender.cs
@@ -643,7 +643,7 @@ namespace log4net.Appender
 		/// <returns>An <see cref="IDbConnection"/> instance with a valid connection string.</returns>
 		protected virtual IDbConnection CreateConnection(Type connectionType, string connectionString)
 		{
-			IDbConnection connection = (IDbConnection)Activator.CreateInstance(connectionType);
+			IDbConnection connection = ObjectFactory.Create<IDbConnection>(connectionType);
 			connection.ConnectionString = connectionString;
 			return connection;
 		}
--- a/src/log4net/Config/PluginAttribute.cs
+++ b/src/log4net/Config/PluginAttribute.cs
@@ -23,9 +23,7 @@
 using System;
 
 using log4net.Core;
-#if !NETSTANDARD1_3
 using log4net.Util;
-#endif
 using log4net.Plugin;
 
 namespace log4net.Config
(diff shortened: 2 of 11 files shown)


NicholasNoise:
@fluffynuts 
I've used custom `PatternConverter`:
```csharp
        /// <inheritdoc />
        protected override void Convert(TextWriter writer, object state)
        {
            if (writer == null)
                throw new ArgumentNullException(nameof(writer));

            string uri = GetUri();
            writer.Write(uri);
        }

        private static string GetUri()
        {
#if NETFRAMEWORK
            try
            {
                var request = HttpContext.Current?.Request;
                return request?.Url.AbsoluteUri;
            }
            catch (Exception)
            {
                return null;
            }
#elif NETSTANDARD
            return null;
#endif
        }
```
But now I'm not able to use it, cuz there are no static `HttpContext` for netstandard2.0 \ netcoreapp, so we need to inject `IHttpContextAccessor` to support this feature.

NicholasNoise:
@fluffynuts 
Pardon me for pushing it, but could you review this one?

fluffynuts:
apologies @NicholasNoise I did look at this in december, but didn't finish up

I was trying to understand why this service-locator has to exist within log4net -- if this is for custom appenders, surely that could be set up in the consuming project and used from within the appender? Even if the custom appender is a nuget package, the service-locator could live in that package.

I don't have a problem with the implementation per se, I'm just wondering why it needs to be in log4net?

NicholasNoise:
>  if this is for custom appenders, surely that could be set up in the consuming project and used from within the appender?

It is easier to use a build-in feature so (any) consuming project could extend or override default behavior. Those changes add the only way to override object creation.

> I don't have a problem with the implementation per se, I'm just wondering why it needs to be in log4net?

You are right, this specific case is not worth it, but in common it makes sense.

fluffynuts:
I'm not sold yet; right now, it adds complexity and risk: it's a way for third-party code to intercept logging when that third-party code takes a guess that an appender uses some built-in service-locator logic.

I'm on the fence. I have to bear in mind that this is not some rinky-dink package like one of my personal ones: where some of mine have had about 300k downloads _for all time_, log4net has garnered just shy of 1/2 a million downloads _since the last release alone_. Changes here affect a lot of people and getting new changes (ie fixes) out takes time because of the processes involved with the Apache org. So far, I can't see this solving a necessary problem. It's not that there's something wrong with the code - quite the opposite - but rather that there's nothing that's specifically broke in log4net _per se_. Your requirements can be met through your own service-locator implementation; even if this was for publishing a third-party appender nuget library, that library could host the service-locator itself, limiting the scope of possible effect by nefarious actors.

Perhaps we'll revisit this some day and I'll be plainly wrong and you'll be able to say "I told you so".

PR #73, opened 2021-02-27
adds ActivatorProvider to allow custom type activation (Resolves LOG4NET-565)

tschettler:
This refactors type activation into an `ActivationProvider` class, which provides wrapper methods for the `IActivator` interface. The default `IActivator` implementation (`DefaultActivator`) derives from the current logic which uses `Activator.CreateInstance`. Clients of log4net can set the IActivator instance to a custom implementation if they so choose. For an example use case, this will allow for dependency injection in custom appenders and pattern converters.

I researched how to do dependency injection in custom log4net appenders and pattern converters without making changes to log4net itself and it proves to be increasingly difficult, since it would require custom classes in cases where Activator.CreateInstance is used. I tried to keep my changes to a bare minimum, as I recognize this project is one of the most popular logging frameworks in .NET and any changes must be heavily scrutinized.

I also recognize that there have been other attempts to add this capability to log4net (#26, #72), but this implementation is merely a refactoring of the existing `Activator.CreateInstance` logic. With all due respect to the authors of the previous implementation attempts, this implementation provides one other key component that the other attempts were missing. The `IsTypeConstructible` method of `XmlHierarchyConfigurator` has been refactored into a `CanCreateInstance` method on the `IActivator` interface. This is necessary to allow custom activation of types without a default constructor.

As this is my first contribution to this project, any feedback is definitely appreciated! 

Thanks!

Travis Schettler

### DIFF
--- a/src/log4net.Tests/Appender/CountingAppenderWithDependency.cs
+++ b/src/log4net.Tests/Appender/CountingAppenderWithDependency.cs
@@ -1,0 +1,25 @@
+﻿using log4net.Appender;
+using log4net.Core;
+
+namespace log4net.Tests.Appender
+{
+    public class CountingContext
+    {
+        public int Count { get; set; }
+    }
+
+    public class CountingAppenderWithDependency : AppenderSkeleton
+    {
+        public CountingAppenderWithDependency(CountingContext context)
+        {
+            Context = context;
+        }
+
+        protected CountingContext Context { get; }
+
+        protected override void Append(LoggingEvent loggingEvent)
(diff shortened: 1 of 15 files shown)


fluffynuts:
@tschettler I think next steps would be to raise a PR with a mvp implmentation of the described pattern in log4net. I don't know when I'll get time to look at an implementation, so if you want to crack on, go for it.

tschettler:
Sounds good, I’d be up for that, thanks!

PR #81, opened 2022-03-18
Enforcing TLS 1.2

BipulRaman:
Enforce TLS 1.2 in PowerShell Scrips as Microsoft is deprecating all old TLS versions


KyleSchoonover (approved):
LGTM

PR #90, opened 2022-08-02
Fix XmlLayoutSchemaLog4j exception thrown in FormatXml

odahcam:
If namespaces writing is disabled it will throw and in some situations produce a broken XML.


fluffynuts:
There is an update in 2.0.15 to handle xml namespacing. Please check it out.

odahcam:
Rrrright, thanks.

PR #91, opened 2022-08-08
SystemInfo.EntryAssemblyLocation writable to adjust location for config-files in unit test projects

FreeAndNil:
We would like to make SystemInfo.EntryAssemblyLocation writable so that we can adjust its value for unit test projects.
.net core sets "testhost.dll" as EntryAssembly.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -243,3 +243,4 @@ buildreports
 vs_buildtools.exe
 dotnetfx35.exe
 *.exe
+/src/Binaries/*
--- a/src/log4net/Util/SystemInfo.cs
+++ b/src/log4net/Util/SystemInfo.cs
@@ -59,7 +59,7 @@ namespace log4net.Util
 		/// Only static methods are exposed from this type.
 		/// </para>
 		/// </remarks>
-		private SystemInfo() 
+		private SystemInfo()
 		{
 		}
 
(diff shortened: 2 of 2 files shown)


fluffynuts (approved):
looks fine to me, tho I can see an existing issue in the default block - Assembly.GetEntryAssembly() can return null; I'll merge and fix in master

FreeAndNil:
Thank you very much.

fluffynuts:
I'm not sure when this will land in a release; I'd like to work around an issue introduced recently which fixes something for pqsql but potentially breaks something for mssql (tho proper config of the latter works around it). Perhaps with that - but I need to set up a proper test scenario first.

FreeAndNil:
No problem. In the meantime I can release a version to our internal Nuget repo.

smx-smx:
Just chiming in to say that this also manifests when hosting .NET in C/C++ (via nethost/fxhost) through my .NET loader (https://github.com/smx-smx/EzDotnet)
If the loaded assembly uses log4net, `Logger.Setup` will fail and logging will not work at all.

This dirty hack works around it, but it's not portable and it can fail when debugging: https://gist.github.com/smx-smx/dc548226bcae16b57f9e6a44ca9d5f99

So i'm also looking forward to the next release 🙂

PR #92, opened 2022-08-08
New locking model for single log file in native applications with managed parts

FreeAndNil:
We have a native application (Delphi) which loads .net dlls at runtime.
With this adjustment we can (optionally) share the log file between the native and the managed part of the application.

### DIFF
--- a/src/log4net/Appender/FileAppender.cs
+++ b/src/log4net/Appender/FileAppender.cs
@@ -160,7 +160,7 @@ namespace log4net.Appender
 				}
 
 #if !NETCR && !NETSTANDARD1_3
-				private LockStateException(SerializationInfo info, StreamingContext context) : base(info, context) 
+				private LockStateException(SerializationInfo info, StreamingContext context) : base(info, context)
 				{
 				}
 #endif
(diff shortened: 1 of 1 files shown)


fluffynuts (approved):
looks ok to me, tho I'd really prefer string interpolation instead of concatenation - I'll merge in and fix up myself

snakefoot:
Just curious how does NoLock work when having two file-handles to the same file ?

Forexample:
- File-handle-1 writes 10 bytes, and moves its file-position 10 byte forward
- File-handle-2 writes 20 bytes and moves its file-position 20 bytes forward

Will File-Handle-2 not overwrite the contents just written by File-Handle-1 ?

With the following code:
```c#
 using var fileStream1 = new FileStream(@"C:\Temp\Hello.txt", FileMode.Append, FileAccess.Write, FileShare.ReadWrite, 1, FileOptions.None);
 using var fileStream2 = new FileStream(@"C:\Temp\Hello.txt", FileMode.Append, FileAccess.Write, FileShare.ReadWrite, 1, FileOptions.None);
 fileStream1.Write(System.Text.Encoding.UTF8.GetBytes("Hello"));
 fileStream2.Write(System.Text.Encoding.UTF8.GetBytes("Goodbye World"));
 fileStream1.Write(System.Text.Encoding.UTF8.GetBytes("World"));
```
Then the contents of the output-file becomes this:
```
GoodbWorldrld
```

PR #95, opened 2022-10-24
Defer creation of locking model to ActivateOptions()

FreeAndNil:
The creation of the locking model instance is now deferred to ActivateOptions() so that defaultLockingModelType has an effect.
+ Tabify all appender code files (were inconsistent)
+ include pdb file in nuget package fpr better debugging experience

### DIFF
--- a/src/log4net/Appender/FileAppender.cs
+++ b/src/log4net/Appender/FileAppender.cs
@@ -1657,7 +1657,7 @@ namespace log4net.Appender
         /// <summary>
         /// The locking model to use
         /// </summary>
-        private FileAppender.LockingModelBase m_lockingModel = new FileAppender.ExclusiveLock();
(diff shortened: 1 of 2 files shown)


fluffynuts:
@FreeAndNil most of the changes here are whitespace, except:
1. not setting `m_lockingModel` at line 1660
2. working around pdb files not being in the nuget package

I'm not 100% sure we want to expand the package size with the pdbs, considering the audience size of log4net. I'm open to be convinced here though (personally, since I use Rider, I can decompile any external sources enough to debug with, so pdbs have been of limited value). Please revert all the other files so that the impact is kept smaller.

FreeAndNil:
Hi @fluffynuts,

i've reverted all the spacing changes.

Regards

fluffynuts (approved):

PR #101, opened 2023-04-26
Add Peek() method to ThreadContextStack and LogicalThreadContextStack.

andreycha:
**Change**
Add `Peek()` method to `ThreadContextStack` and `LogicalThreadContextStack` to be able to get the most recent value from the context without removing it.

**Motivation**
In the scope of [ElasticCommonSchema .NET integration for log4net](https://github.com/elastic/ecs-dotnet/tree/main/src/Elastic.CommonSchema.Log4net) log entry properties are being read and written to ECS format. In case of context stacks [current implementation](https://github.com/elastic/ecs-dotnet/blob/main/src/Elastic.CommonSchema.Log4net/LoggingEventConverter.cs#L85) takes string representation of the stack. It works fine when stack (i.e., property) has a single value, but there are also cases when stack contains multiple values (same or different ones). With the proposed changes it will be able to take the most recent value, in the context of which the current log entry is being written.

Consider following scenario:

```cs
using (LogicalThreadContext.Stacks["Id"].Push("123"))
{
    log.Info("Log 1"); // currently has "123" in the log metadata

    using (LogicalThreadContext.Stacks["Id"].Push("456"))
    {
        log.Info("Log 2"); // currently has "123 456" in the log metadata; should have "456" instead

        using (LogicalThreadContext.Stacks["Id"].Push("789"))
        {
            log.Info("Log 3"); // currently has "123 456 789" in the log metadata; should have "789" instead
        }
    }
}
```

**Alternative solutions considered**
An alternative approach might be to just pop the most recent value and then push it again onto the stack. Problem here is that it creates a new scope which will not be closed as no one will call `Dispose()` on the returned result from `Push()`.

### DIFF
--- a/src/log4net/Util/LogicalThreadContextStack.cs
+++ b/src/log4net/Util/LogicalThreadContextStack.cs
@@ -198,15 +198,35 @@ namespace log4net.Util
 			return new AutoPopStackFrame(contextStack, stack.Count - 1);
 		}
 
-		#endregion Public Methods
+        /// <summary>
+        /// Returns the top context from this stack.
+        /// </summary>
+        /// <returns>The message in the context from the top of this stack.</returns>
+        /// <remarks>
+        /// <para>
+        /// Returns the top context from this stack. If this stack is empty then an
+        /// empty string (not <see langword="null"/>) is returned.
+        /// </para>
+        /// </remarks>
+        public string Peek()
+        {
+            Stack stack = m_stack;
+            if (stack.Count > 0)
+            {
+                return ((StackFrame)stack.Peek()).Message;
+            }
+            return "";
+        }
 
-		#region Internal Methods
+        #endregion Public Methods
 
-		/// <summary>
-		/// Gets the current context information for this stack.
-		/// </summary>
-		/// <returns>The current context information.</returns>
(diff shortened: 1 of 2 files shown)


andreycha:
Hi @fluffynuts , could you please check this PR? Thanks!

fluffynuts (approved):
LGTM

I can't make any promises about when this gets out live - I haven't had a lot of time to put into this in a while ): I'm currently formulating a plan to do beta / prerelease packages with less release friction so that perhaps at least things can get out quicker, but I have to speak with more venerable Apache people on what's process would be acceptable.

andreycha:
Thanks! Is there a way to somehow get notified when a new release comes out?

fluffynuts:
@andreycha I guess you could sub to apache logging email lists; but I really want to figure out a way to do beta releases, and I doubt they'd go through that, so :shrug: I haven't figured this out yet ): I have this tension between "actual releases are a major mission with a lot of dead time whilst I wait for a bunch of Java people to vote on a dotnet thing that they have no context about" and "I just want to release fixes" )':

andreycha:
Yeah, I can relate. Anyway, looking forward to the new release!

PR #105, opened 2024-01-24
Fix CodeQL Password in configuration file Rule False Positives

justinmichaels:
CodeQL is falsly identifying the documentation examples of SQL Server connection strings violating [CodeQL: c# Password in configuration file Rule](https://codeql.github.com/codeql-query-help/csharp/cs-password-in-configuration/). Even though it's a false positive, many organizations take stances of not allowing dismissals and would rather see fixes for CodeQL findings. This proposed change will prevent this rule from being tripped in the future and not require organizations to mark these as false positives manually.


FreeAndNil (approved):
lgtm

PR #115, opened 2024-02-29
#111 Dropping support for older runtimes

### LINKED ISSUE #111: Dropping support for older runtimes
Log4net supports a lot of really old runtimes, including net2.0 and net3.5 _CE_ (Compact Edition) which can be notoriously difficult to install on windows (I've been lucky and unlucky - today, it "just worked" without intervention, but previously, I've crafted the included powershell scripts to install .net framework 3.5.

These targets make it impossible (as far as I know, and I've tried) to build on non-windows hosts. However, if we could reduce the targeted runtime list to:
- net462
- netstandard2.0

then we could support all supported versions of .net (at time of writing, the lowest supported netfx version is 4.6.2) because we can use Mono on non-windows hosts to provide the api to compile system functionality against (see [MonoForFramework.targets](https://github.com/apache/logging-log4net/blob/master/src/MonoForFramework.targets))

Apart from quicker builds & testing times, this would be highly beneficial in that it would make contributing to log4net much easier.

On one hand, I'd really like to maintain as much compatibility as possible, but on the other, I'd like to remove hurdles for prospective devs (and myself 😅 )

FreeAndNil:
- changed target frameworks to net462 & netstandard2.0
- added release notes for 3.0.0
- resolves #111


FreeAndNil:
@fluffynuts I've executed steps 1 & 2 of releasing.md

Could you take a look?

Thanks

fluffynuts:
Will do

fluffynuts:
I'm wondering if we shouldn't try get out a 2.x release before this - just release off of master. Perhaps even one more after that (there was an issue with mssql introduced by some update to make pgsql better, to do with when parameterisation is done (I forget the rest of the details)).

At the very least, it would solve the "when will this patch be released" questions. Let's keep this branch around though - I want to simplify build, and you've done the right thing (imo) with the major version bump. If there's some really good reason to revive a 2.x release, we could then also use the last release tag as a starting point. What do you think?

FreeAndNil:
Thats a good Idea.I will separate the release note changes into a separate branch and adjust them to a 2.16.Regards.JanOn 1 Mar 2024 07:10, Davyd McColl ***@***.***> wrote:
I'm wondering if we shouldn't try get out a 2.x release before this - just release off of master. Perhaps even one more after that (there was an issue with mssql introduced by some update to make pgsql better, to do with when parameterisation is done (I forget the rest of the details)).
At the very least, it would solve the "when will this patch be released" questions. Let's keep this branch around though - I want to simplify build, and you've done the right thing (imo) with the major version bump. If there's some really good reason to revive a 2.x release, we could then also use the last release tag as a starting point. What do you think?

src/log4net.sln
```diff
@@ -1,28 +1,10 @@
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 2012
```
fluffynuts, line 2: don't forget to add this license back in again, otherwise someone will just ask you to later 🤣

src/log4net/log4net.csproj
```diff
@@ -66,6 +37,20 @@
...
+    <Copyright>Copyright 2004-2024 The Apache Software Foundation</Copyright>
+    <DevelopmentDependency>false</DevelopmentDependency>
+    <GeneratePackageOnBuild>false</GeneratePackageOnBuild>
+    <PackageLicenseExpression>Apache-2.0</PackageLicenseExpression>
+    <PackageProjectUrl>https://logging.apache.org/log4net/</PackageProjectUrl>
+    <PackageIcon>package-icon.png</PackageIcon>
+    <PackageTags>logging log tracing logfiles</PackageTags>
+    <RepositoryType>git</RepositoryType>
+    <RepositoryUrl>https://github.com/apache/logging-log4net</RepositoryUrl>
+    <Owners>Apache Logging Project, Jiří Činčura</Owners>
```
fluffynuts, line 51: I think this should be changed to "Apache Logging Project" - Jiří Činčura is github user cincuranet and I can neither find github issues/prs from that user nor can I find commits with that name or username. The project is owned by the Apache Logging Project, afaik.

fluffynuts (changes requested):

src/log4net/log4net.csproj
FreeAndNil: Jiří Činčura was the first one to push log4net to nuget.org
I've changed it.

src/log4net.sln
FreeAndNil: check

.editorconfig
```diff
@@ -0,0 +1,2 @@
+indent_style = space
+indent_size = 2
```
fluffynuts, line 2: I think I saw this last time too - please change to 4, simply because that's the community standard (and what existing files would be indented to)

Always choose the community standard for style - it enables easier access & contribution by others.

fluffynuts (changes requested):

.editorconfig
FreeAndNil: Thanks for the hint. Adjusted.

src/log4net/Repository/Hierarchy/Hierarchy.cs
```diff
@@ -317,50 +203,23 @@ public override ILogger Exists(string name)
...
-        {
-          if (node is Logger)
-          {
-            loggers.Add(node);
-          }
-        }
-        return (Logger[])loggers.ToArray(typeof(Logger));
-      }
+      // loggers are Logger objects as there might be some ProvisionNodes as well.
+      return loggers.Values.OfType<Logger>().ToArray();
```
snakefoot, line 207: Notice that the `Values`-property on ConcurrentDictionary is dangerous (locks the entire collection). Maybe do this instead:
```c#
return loggers.Select(l => l.Value).OfType<Logger>().ToArray();
```

FreeAndNil: @snakefoot thanks for the hint - I was not aware of this.

ppkarwasz:
Since `3.0.0` was released from this branch, shouldn't you merge this PR?

fluffynuts (approved):

Issue #118, opened 2024-03-08
NullRefException from shutdown in 2.0.16


erikmav:
[This commit](https://github.com/apache/logging-log4net/commit/b2ef9f6ced9d755844f13586f189a814e20c0c91) introduced after 2.0.15 results in an NRE during some shutdown cases where m_lockingModel is null during Close/Dispose.

The simplistic fix is to add a check to FileAppender. The better fix is to enable nullability on FileAppender, RollingFileAppender, and their unit tests, and fix all nullability problems.

A fix is needed as a 2.0.17 package in addition to the proposed 3.0 package.

erikmav:
@FreeAndNil

FreeAndNil:
Thanks for reporting.
I will take a look.

erikmav:
I'm putting together a large PR with the nullability setting change. It's a lot to review so for faster release it's probably better to fix the dispose case as a quick fix then ship 2.0.17.

FreeAndNil:
@erikmav I agree. Please rebase/start your nullability PR on Feature/111-Dropping-support-for-older-runtimes.
Will you provide the quick fix or should I do?

erikmav:
Can you add the quick fix? I'm heads-down on the large PR.

erikmav:
@FreeAndNil Small and large PRs posted.

FreeAndNil:
Will be fixed in 2.0.17 (2024-03-22).

andreycha:
Hi @FreeAndNil any ETA when 2.0.17 stable version is going to be released? I need some changed from 2.0.16 but don't want to upgrade without this hotfix.

FreeAndNil:
@andreycha definitely next week.

FreeAndNil:
@andreycha
https://github.com/apache/logging-log4net/releases/tag/rel%2F2.0.17
https://www.nuget.org/packages/log4net/2.0.17

andreycha:
@FreeAndNil thanks a lot!

PR #119, opened 2024-03-08
Enable nullable for FileAppender, RollingFileAppender, and unit tests

erikmav:
For #118 .
And fix null issues. Also includes adding friend reference for unit tests to remove reflection code across unit tests for type safety, removing ArrayList usage in favor or strongly typed lists, and a few opportunistic auto-properties and other modernizations. Doing this required restoring strong name for log4net.Tests, which then led to converting 3 NExpect uses to Assert and removing the NExpect reference, as it is not strong-named and net462 requires strong named assemblies to use only strong named assemblies.


src/log4net/Appender/RollingFileAppender.cs
```diff
@@ -1376,19 +1327,22 @@ protected void RollOverSize()
...
-      LogLog.Debug(declaringType, "maxSizeRollBackups [" + m_maxSizeRollBackups + "]");
+      LogLog.Debug(declaringType, "maxSizeRollBackups [" + MaxSizeRollBackups + "]");
       LogLog.Debug(declaringType, "curSizeRollBackups [" + m_curSizeRollBackups + "]");
-      LogLog.Debug(declaringType, "countDirection [" + m_countDirection + "]");
+      LogLog.Debug(declaringType, "countDirection [" + CountDirection + "]");
 
-      RollOverRenameFiles(File);
+      if (File is not null)
+      {
+          RollOverRenameFiles(File);
```
FreeAndNil, line 1336:
```suggestion
        RollOverRenameFiles(File);
```

FreeAndNil (changes requested):
Thanks. I only found a few little things.

src/log4net/Appender/FileAppender.cs
```diff
@@ -779,16 +782,26 @@ public override void ActivateOptions()
       {
         if (m_mutex == null)
         {
-          string mutexFriendlyFilename = CurrentAppender.File
-              .Replace("\\", "_")
-              .Replace(":", "_")
-              .Replace("/", "_");
+          if (CurrentAppender is not null)
```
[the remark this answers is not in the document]
erikmav: Examined but prefer to keep this one, the outer if has no else{Error()} because a null CurrentAppender can be ignored, whereas the inner if has an error output.

src/log4net/Appender/RollingFileAppender.cs
erikmav: My 4-spaces settings were fighting the 2-space code. Added a .editorconfig in the next push.

FreeAndNil (approved):
Thanks a lot.

FreeAndNil:
@erikmav I've created a new branch and PR because I couldn't resolve the conflicts in the old one.
#119

Issue #124, opened 2024-03-13
Add support for nullable annotations


FreeAndNil:
Public classes and interfaces should get nullable annotations so that callers will be warned when passing null values.

#### #129 - Part 4 of nullable and modernization
- interface `log4net.Appender.AppenderCollection.IAppenderCollectionEnumerator` is now obsolete and must be replaced by `IEnumerator<IAppender>`
- `log4net.Appender.MemoryAppender.m_eventsList` (protected field) was changed from `ArrayList` to `List<LoggingEvent>` 
  - for getting `m_eventsList.SyncRoot` you need to cast it: `((ICollection)m_eventsList).SyncRoot`

PR #126, opened 2024-03-15
Nullability and modernization updates, part 3

erikmav:
For #124

### DIFF
--- a/src/log4net/Core/LoggingEvent.cs
+++ b/src/log4net/Core/LoggingEvent.cs
@@ -19,10 +19,12 @@
 
 using System;
 using System.Collections.Generic;
+using System.Globalization;
 using System.IO;
 using System.Runtime.Serialization;
 using System.Security;
 using System.Security.Principal;
+using System.Threading;
 using log4net.Util;
 using log4net.Repository;
 
@@ -676,12 +678,12 @@ namespace log4net.Core
     {
       get
       {
-        if (m_data.ThreadName == null && this.m_cacheUpdatable)
+        if (m_data.ThreadName == null && m_cacheUpdatable)
         {
           // '.NET ThreadPool Worker' appears as a default thread name in the .NET 6-7 thread pool.
           // '.NET TP Worker' is the default thread name in the .NET 8+ thread pool.
           // Prefer the numeric thread ID instead.
-          string threadName = System.Threading.Thread.CurrentThread.Name;
+          string threadName = Thread.CurrentThread.Name;
           if (!string.IsNullOrEmpty(threadName) && threadName != ".NET TP Worker" && threadName != ".NET ThreadPool Worker")
           {
             m_data.ThreadName = threadName;
@@ -693,9 +695,7 @@ namespace log4net.Core
             // current thread. (Why don't Threads know their own ID?)
             try
             {
-              m_data.ThreadName =
-                  SystemInfo.CurrentThreadId.ToString(System.Globalization.NumberFormatInfo
-                      .InvariantInfo);
+              m_data.ThreadName = SystemInfo.CurrentThreadId.ToString(NumberFormatInfo.InvariantInfo);
             }
             catch (SecurityException)
             {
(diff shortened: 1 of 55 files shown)


erikmav:
For #124

erikmav:
@FreeAndNil Important to closely review `ReadOnlyPropertiesDictionary` and `PropertiesDictionary`. I selected the most backward compatible option, which is to coerce the `this[]` getter to avoid throwing KeyNotFoundException like Dictionary<> would (including a new test for the behavior). Also note that to avoid a double-indirection it would be ideal to change both to structs.

FreeAndNil:
Thanks. I will review on Monday.

src/log4net/Util/ThreadContextStack.cs
```diff
@@ -223,7 +208,7 @@ internal Stack InternalStack
     /// Gets the current context information for this stack.
     /// </para>
     /// </remarks>
-    public override string ToString()
+    public override string? ToString()
```
FreeAndNil, line 211:
```suggestion
    public override string? ToString() => GetFullMessage();
```

src/log4net/Util/ThreadContextStack.cs
```diff
@@ -239,7 +224,7 @@ public override string ToString()
     /// </remarks>
     object IFixingRequired.GetFixedObject()
```
FreeAndNil, line 225:
```suggestion
    object IFixingRequired.GetFixedObject() => GetFullMessage() ?? string.Empty;
```

src/log4net/Util/TypeConverters/ConverterRegistry.cs
```diff
@@ -208,24 +180,19 @@ public static IConvertFrom GetConvertFrom(Type destinationType)
...
       foreach (var attribute in attributes)
       {
-        var tcAttr = attribute as TypeConverterAttribute;
-        if (tcAttr != null)
+        if (attribute is TypeConverterAttribute tcAttr)
         {
-          var converterType = SystemInfo.GetTypeFromString(destinationType, tcAttr.ConverterTypeName, false, true);
-          return CreateConverterInstance(converterType);
+          Type? converterType = SystemInfo.GetTypeFromString(destinationType, tcAttr.ConverterTypeName, false, true);
+          if (converterType is not null)
```
FreeAndNil, line 192:
```suggestion
          if (SystemInfo.GetTypeFromString(destinationType, tcAttr.ConverterTypeName, false, true) is Type converterType)
```

FreeAndNil (changes requested):
Thanks. I've made a few suggestions.

src/log4net/Core/LoggingEvent.cs
```diff
@@ -826,11 +825,10 @@ private static string TryReadWindowsIdentityUserName()
         {
           try
           {
-            if (System.Threading.Thread.CurrentPrincipal is not null &&
-                System.Threading.Thread.CurrentPrincipal.Identity is not null &&
-                System.Threading.Thread.CurrentPrincipal.Identity.Name is not null)
+            if (Thread.CurrentPrincipal is not null &&
```
FreeAndNil, line 828: I think we can shorten this even more:
```suggestion
            if (Thread.CurrentPrincipal?.Identity?.Name is string name)
```

FreeAndNil (changes requested):
@erikmav one tiny thing left.

FreeAndNil:
> Also note that to avoid a double-indirection it would be ideal to change both to structs.
@erikmav can you explain how you would implement this? Maybe as separate issue?

erikmav:
Could be done under another issue. The change is to change from class to struct (or readonly struct, depending) for ReadOnlyPropertiesDictionary, then derive PropertiesDictionary from that as a second struct. I wouldn't do it in this PR either way, it's mainly an optimization.

FreeAndNil (approved):
Thanks a lot.

Issue #127, opened 2024-03-17
Enable building log4net with docker container


FreeAndNil:
Provide the following convenience for PMC members to review the project:
```
git checkout <release-tag>
docker build -t log4net-builder .
docker run log4net-builder
```
That is, a convenience to
build a Docker image containing all necessary dependencies to be able to build the project
an entry point to build the project using the created Docker image.

Proposed by @vy

PR #128, opened 2024-03-17
Enable building log4net with docker container #127

FreeAndNil:
Enable building log4net with docker container #127


Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+  && chmod +x ./dotnet-install.sh \
+  && ./dotnet-install.sh --channel 8.0
+ENV DOTNET_NOLOGO=true
+ENV DOTNET_SKIP_FIRST_TIME_EXPERIENCE=true
+ENV DOTNET_ROOT=/root/.dotnet
+ENV PATH="$PATH:$DOTNET_ROOT:$DOTNET_ROOT/tools"
+  
+ADD . /logging-log4net
+
+CMD ["/logging-log4net/docker-build.sh"]
```
vy, line 35: This doesn't work for me:

```
docker: Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: exec: "/logging-log4net/docker-build.sh": permission denied: unknown.
```

I guess the executable bit is not reflected. I am not gonna Google to try to figure out why, since I know you can do that too. I will appreciate it if you can fix this. In the worst case:

```suggestion
CMD ["sh", "/logging-log4net/docker-build.sh"]
```

vy (changes requested):
Great start @FreeAndNil! Thanks so much. I would appreciate it if you can implement the following too:

1. `BUILDING.md`, `building.xml`, `docker-build.sh`, etc. all mention different instructions. I just want to _"build the sources and run tests"_. Could you provide me a single entry point (shell script, etc.) for that and document it wherever necessary, please? Contrast this with the following: https://logging.apache.org/logging-parent/latest/#development These instructions here apply to *all* Maven-based Logging Services subprojects; Log4j, Log4j Scala, Log4j Kotlin, Log4j Tools, etc. There is no discrepancy. This is what CI (GitHub Actions), more or less, uses too. I prefer such a universal build process.
1. Please treat me like a monkey who doesn't know what to type into the terminal. What are the host requirements? (`docker` command line and internet access?) Which commands do I exactly need to run?  etc.
1. All `docker` commands require network interaction, hence they won't work without `--network host`.

Dockerfile
FreeAndNil: Thanks for the hint.
The chmod +x was missing.
For whatever reason this worked on my Docker Desktop install.

FreeAndNil:
@vy I've updated the dockerfile, the scripts and https://github.com/apache/logging-log4net/blob/Feature/127-docker-build/doc/BUILDING.md

Please take a look.

@fluffynuts what is your opinion?

fluffynuts:
lgtm, tho I'm no docker expert, so as long as it works, which I assume it does (I'm nowhere near a PC rn), then I give it 👍👍

Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+  && chmod +x ./dotnet-install.sh \
+  && ./dotnet-install.sh --channel 8.0
+ENV DOTNET_NOLOGO=true
+ENV DOTNET_SKIP_FIRST_TIME_EXPERIENCE=true
+ENV DOTNET_ROOT=/root/.dotnet
+ENV PATH="$PATH:$DOTNET_ROOT:$DOTNET_ROOT/tools"
+  
+ADD . /logging-log4net
+RUN chmod +x /logging-log4net/docker-build/docker-build.sh
+ADD ./docker-build/.bashrc /root/.bashrc
```
vy, line 35: @FreeAndNil, sorry for nit picking, but given the extremely simple contents of `.bashrc` and `docker-build.sh`, I would prefer you remove those files and instead inline their logic here via `CMD` or something similar. The complexity is already mind blowing, let's not bloat it more by introducing extra layers of delegation.

package.json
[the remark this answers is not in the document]
FreeAndNil: Good point. Changed.

Dockerfile
FreeAndNil: You're right. I changed it.

vy (approved):
@FreeAndNil, looks great! :star_struck:
I tested the `Dockerfile` and it works like a charm.
Thanks so much for the awesome work! :bow:

PR #132, opened 2024-03-27
Nullable and modernization part 6

erikmav:
Adjusted nullability on some classes after applying the 3.0 binary to two existing projects that have nullable enabled. Changed nullable warnings to errors.

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -3,5 +3,6 @@
         <UseSharedCompilation>true</UseSharedCompilation>
         <LangVersion>latest</LangVersion>
         <Nullable>Enable</Nullable>
+        <WarningsAsErrors>nullable</WarningsAsErrors>
     </PropertyGroup>
 </Project>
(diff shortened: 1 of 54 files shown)


erikmav:
For #124

FreeAndNil (approved):
Thanks a lot.
I will test the API with our libraries.

erikmav:
@FreeAndNil  the ugliest adaptations I had to make with the new binary:
* `Level?` on LoggingEvent and its internal data structure required adding a bunch of null checks, especially where there are comparisons of levels. Lots of these in 2 repos. It would be ideal to make that non-nullable and ensure it's set to some default like `Level.Off` or similar.
* `Repository?` on Logger required adding null checks. Generally speaking this is never null in practice, but I don't see what default to set it to.

erikmav:
@FreeAndNil I just tried setting LoggingEvent.Level to non-nullable and set to Level.All. Problem is that null is converted into different policies by different appenders (Debug, Info, or All) so a blanket All policy does not work. Seems like it needs to be left nullable for back compat.

PR #136, opened 2024-03-28
Nullable: Eliminate nullable in one place, and clarify the meaning of null levels in doc comments

erikmav:
For #124


FreeAndNil (approved):
Thanks.

PR #137, opened 2024-03-28
#124 LoggingEvent.Level not nullable

FreeAndNil:
#124 LoggingEvent.Level not nullable
@erikmav what do you think about this approach?


erikmav:
In the 2.x codebase a null level was handled as an allowed option, and some formatters use default behaviors on a null value (search for `level is null` across the codebase). IOW this is probably a back compat problem where we might break people. Maybe that's OK in v3... which means we need `if (level is null) { throw new ArgumentNullException(nameof(level)); }` anywhere we're receiving a level from the user.

PR #140, opened 2024-04-05
3.0.0 preview.1

FreeAndNil:


erikmav (approved):

src/build-preview.ps1
```diff
@@ -0,0 +1 @@
+dotnet build -c Release '-p:GeneratePackages=true;VersionSuffix=preview.1' ./log4net/log4net.csproj
```
fluffynuts, line 1: zarro can do this, though not with a set suffix (it will generate one from the current date/time and sha) - but I should add support for named versions. This is fine for now (:

fluffynuts (approved):

src/log4net/log4net.csproj
```diff
@@ -63,10 +62,10 @@
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)'=='Debug' ">
     <DefineConstants>TRACE;DEBUG;$(DefineConstants)</DefineConstants>
+    <GeneratePackageOnBuild>$(GeneratePackages)</GeneratePackageOnBuild>
```
fluffynuts, line 65: why auto-generate packages just for debug builds? this will slow down any TDD effort, btw, as you'd be building packages at every point in the red-green-refactor cycle, but I'm assuming there's a good reason?

FreeAndNil: Please see line 69.
For Debug-Builds you can turn on package generation via "GeneratePackages" (default: false).
For Release-Builds they are always on (assuming you do TDD with Debug binaries ;-)

Issue #142, opened 2024-04-18
.NET8.0 XML Exception Problem


SwEngine:
I was using the package in .net7.0 without problems. However, after updating to .net8.0, my program throws error. When I unchecked the trimming unused code option, there was no error, but it increases output file size very much. How can I solve it?

FreeAndNil:
Hi @SwEngine,

can you please be more specific?
What exact error so you get when you do **what**?

SwEngine:
I use only normal logging processes. When releasing the executable, I checked the trimming unused code option, it gives the error below,
However, when I unchecked the trimming unused code option, it does not give any error.

`log4net:ERROR Exception while reading ConfigurationSettings. Check your .config file is well formed XML.
System.Configuration.ConfigurationErrorsException: Configuration system failed to initialize
 ---> System.MissingMethodException: Cannot dynamically create an instance of type 'System.Configuration.ClientConfigurationHost'. Reason: No parameterless constructor defined.
   at System.RuntimeType.ActivatorCache..ctor(RuntimeType)
   at System.RuntimeType.CreateInstanceDefaultCtor(Boolean, Boolean)
   at System.Configuration.Internal.ConfigSystem.System.Configuration.Internal.IConfigSystem.Init(Type, Object[] )
   at System.Configuration.ClientConfigurationSystem..ctor()
   at System.Configuration.ConfigurationManager.EnsureConfigurationSystem()
   --- End of inner exception stack trace ---
   at System.Configuration.ConfigurationManager.PrepareConfigSystem()
   at System.Configuration.ConfigurationManager.GetSection(String)
   at System.Configuration.ConfigurationManager.get_AppSettings()
   at log4net.Util.SystemInfo.GetAppSetting(String)
log4net:ERROR Exception while reading ConfigurationSettings. Check your .config file is well formed XML.
System.Configuration.ConfigurationErrorsException: Configuration system failed to initialize
 ---> System.MissingMethodException: Cannot dynamically create an instance of type 'System.Configuration.ClientConfigurationHost'. Reason: No parameterless constructor defined.
   at System.RuntimeType.ActivatorCache..ctor(RuntimeType)
   at System.RuntimeType.CreateInstanceDefaultCtor(Boolean, Boolean)
   at System.Configuration.Internal.ConfigSystem.System.Configuration.Internal.IConfigSystem.Init(Type, Object[] )
   at System.Configuration.ClientConfigurationSystem..ctor()
[...]

PR #143, opened 2024-04-18
#142 opt-out of trimming until it works

FreeAndNil:
#142 opt-out of trimming until it works
see also https://github.com/dotnet/runtime/pull/49085 & https://github.com/dotnet/runtime/issues/49062


FreeAndNil:
Abandoned until we get feedback from #142 or #147

Issue #156, opened 2024-07-26
Regression: Creating nested loggers in reverse order fails in 3.0.0-preview.1


FreeAndNil:
When creating nested loggers in reverse order

`var child = LogManager.GetLogger("A.B.C");`
`var parent = LogManager.GetLogger("A.B")`

you get

`  Message: `
`    System.InvalidCastException : Unable to cast object of type 'log4net.Repository.Hierarchy.ProvisionNode' to type 'log4net.Repository.Hierarchy.Logger'.`

`  Stack Trace: `
`    <>c__DisplayClass32_0.<GetLogger>g__CreateLogger|1(Action`1 extraInit) line 688`
`    Hierarchy.GetLogger(String name, ILoggerFactory factory) line 679`
`    Hierarchy.GetLogger(String name) line 317`

src/log4net/Repository/Hierarchy/Hierarchy.cs in b5f6ea2df90372cbe3bc679e6fffa07b984fdb6e #126

FreeAndNil:
Fixed in 3.0.0-preview.2

PR #157, opened 2024-07-26
#156 fixed Regression when creating nested loggers in reverse order

FreeAndNil:
#156 fixed Regression when creating nested loggers in reverse order


fluffynuts (changes requested):

src/log4net/Repository/Hierarchy/Hierarchy.cs
```diff
@@ -660,43 +660,62 @@ public Logger GetLogger(string name, ILoggerFactory factory)
...
       var key = new LoggerKey(name);
 
-      // Synchronize to prevent write conflicts. Read conflicts (in
-      // GetEffectiveLevel() method) are possible only if variable
-      // assignments are non-atomic.
+      while (true)
+      {
+        if (TryCreateLogger(key, factory) is Logger result)
+        {
+          return result;
```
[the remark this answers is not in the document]
FreeAndNil: Thanks for the hint.

src/log4net/Repository/Hierarchy/Hierarchy.cs
```diff
@@ -203,22 +203,22 @@ public Logger Root
     {
       get
       {
-        if (m_root is null)
+        if (rootLogger is null)
         {
           lock (this)
```
fluffynuts, line 208: i know this is the existing code, but `lock(this)` is potentially problematic - something else could lock `this` from the outside; I'd recommend adding a lock object and locking that; probably ok though - it's probably been this way since the beginning.

src/log4net/Repository/Hierarchy/Hierarchy.cs
```diff
@@ -252,12 +249,7 @@ public ILoggerFactory LoggerFactory
     /// </remarks>
     public override ILogger? Exists(string name)
     {
-      if (name is null)
-      {
-        throw new ArgumentNullException(nameof(name));
-      }
-
-      m_ht.TryGetValue(new LoggerKey(name), out object? o);
+      loggers.TryGetValue(new(name.EnsureNotNull()), out object? o);
```
michal-ciechan, line 252: I would have kept `name.EnsureNotNull()` on its own line at the start.

FreeAndNil: inlining EnsureNotNull() is one of the reasons this method exists ;-)

src/log4net/Repository/Hierarchy/Hierarchy.cs
FreeAndNil: Thanks. I resolved it with Interlocked.CompareExchange.

fluffynuts (approved):

Issue #165, opened 2024-08-11
Configuration System Initialization Failure with PublishTrimmed in .NET 8


tmoonlight:
Hi,

I am using log4net in my project running on .NET 8. When I enable the “PublishTrimmed” setting in my project file (< PublishTrimmed >true< /PublishTrimmed >), I consistently receive the following errors during the configuration of log4net:

```

Starting up as a console service host
Service NSPServer started
The NSPServer service is now running, press Control+C to exit.
Another instance of the program is running. It may cause fatal error.
Initializing..
log4net:ERROR Exception while reading ConfigurationSettings. Check your .config file is well formed XML.
System.Configuration.ConfigurationErrorsException: Configuration system failed to initialize
 ---> System.MissingMethodException: Cannot dynamically create an instance of type 'System.Configuration.ClientConfigurationHost'. Reason: No parameterless constructor defined.
   at System.RuntimeType.ActivatorCache..ctor(RuntimeType)
   at System.RuntimeType.CreateInstanceDefaultCtor(Boolean, Boolean)
   at System.Activator.CreateInstance(Type type, Boolean nonPublic, Boolean wrapExceptions)
   at System.Activator.CreateInstance(Type type, Boolean nonPublic)
   at System.Configuration.TypeUtil.CreateInstance(Type)
   at System.Configuration.Internal.ConfigSystem.System.Configuration.Internal.IConfigSystem.Init(Type, Object[] )
   at System.Configuration.ClientConfigurationSystem..ctor()
   at System.Configuration.ConfigurationManager.EnsureConfigurationSystem()
   --- End of inner exception stack trace ---
   at System.Configuration.ConfigurationManager.PrepareConfigSystem()
   at System.Configuration.ConfigurationManager.GetSection(String)
   at System.Configuration.ConfigurationManager.get_AppSettings()
   at log4net.Util.SystemInfo.GetAppSetting(String)
``` 

This issue persists even when I configure log4net purely through code, without using a log4net.config file. The problem does not occur in non-trimmed mode or older versions of the trimming mode (such as .NET Core 3.0).

To assist in diagnosing this issue, I have attached a file containing my program that can reliably reproduce the problem:
https://github.com/tmoonlight/NSmartProxy/releases/download/v1.4.0/nspclient_scd_win_x86_v1.4.0.zip

FreeAndNil:
related to #147 
related to #142

FreeAndNil:
[...]

Issue #147, opened 2024-05-09
Support aot?


YahuiWong:
I publish my  dotnet 8 project by aot mode ,
After that the soft can`t working . and print the err msg 
```
log4net:ERROR Exception while reading ConfigurationSettings. Check your .config file is well formed XML.
System.Configuration.ConfigurationErrorsException: Configuration system failed to initialize
 ---> System.MissingMethodException: No parameterless constructor defined for type 'System.Configuration.ClientConfigurationHost'.
   at System.ActivatorImplementation.CreateInstance(Type, Boolean) + 0x119
   at System.Configuration.Internal.ConfigSystem.System.Configuration.Internal.IConfigSystem.Init(Type, Object[]) + 0x43
   at System.Configuration.ClientConfigurationSystem..ctor() + 0x47
   at System.Configuration.ConfigurationManager.EnsureConfigurationSystem() + 0x56
   --- End of inner exception stack trace ---
   at System.Configuration.ConfigurationManager.PrepareConfigSystem() + 0x35
   at System.Configuration.ConfigurationManager.GetSection(String) + 0x18
   at System.Configuration.ConfigurationManager.get_AppSettings() + 0x11
   at log4net.Util.SystemInfo.GetAppSetting(String) + 0x21
```

YahuiWong:
> @YahuiWong please See #142

Is  https://github.com/apache/logging-log4net/issues/142 have a solution?

FreeAndNil:
@YahuiWong yes, see my comments and #143

FreeAndNil:
@YahuiWong please reopen when you have feedback

YahuiWong:
I have no idea about how to solve it.
Is that I reference log4net  need use dll files not nuget?

FreeAndNil:
> I have no idea about how to solve it.
> Is that I reference log4net  need use dll files not nuget?
> @FreeAndNil 

@YahuiWong yes, you need to build the dll from the sources in the refrenced branch or use the linked binaries in #142

PR #166, opened 2024-08-11
opt-out of trimming until it works #142, #147, #165

### LINKED ISSUE #142: .NET8.0 XML Exception Problem

FreeAndNil:
may fix #142, #147 and #165


FreeAndNil:
Closing as of now.

Issue #169, opened 2024-08-14
System.NullReferenceException when comparing with a null Level


FreeAndNil:
```
  Level? left = null;
  Level? right = null;
  Assert.IsFalse(left < right);
```
leads to

System.NullReferenceException : Object reference not set to an instance of an object.

Stack Trace: 
Level.op_LessThan(Level l, Level r) line 346
LevelTest.LevelCompare() line 37

FreeAndNil:
Fixed in 3.0.0-Preview.3

PR #170, opened 2024-08-14
#169 fixed NullReferenceException when comparing with a null Level

FreeAndNil:
#169 fixed NullReferenceException when comparing with a null Level

### DIFF
--- a/src/log4net/Core/Level.cs
+++ b/src/log4net/Core/Level.cs
@@ -20,523 +20,462 @@
 using System;
 using System.Runtime.Serialization;
 using System.Security;
-using System.Xml.Linq;
 using log4net.Util;
 
-namespace log4net.Core
+namespace log4net.Core;
+
+/// <summary>
+/// Defines the default set of levels recognized by the system.
+/// </summary>
+/// <remarks>
+/// <para>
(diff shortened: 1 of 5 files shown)


src/log4net/Core/Level.cs
```diff
@@ -20,523 +20,468 @@
...
-    /// <param name="info">The <see cref="SerializationInfo" /> that holds the serialized object data.</param>
-    /// <param name="context">The <see cref="StreamingContext" /> that contains contextual information about the source or destination.</param>
-    /// <remarks>
-    /// <para>
-    /// Initializes a new instance of the <see cref="Level" /> class 
-    /// with serialized data.
-    /// </para>
-    /// </remarks>
-    protected Level(SerializationInfo info, StreamingContext context)
+    if (l is null && r is null)
```
fluffynuts, line 367: unnecessary - if both are null, `ReferenceEquals` returns true.

fluffynuts (approved):
only a minor comment about an unnecessary check, but it's not the end of the world, and perhaps clarifies for the reader - it's up to you what you do about it.

src/log4net/Core/Level.cs
FreeAndNil: Thanks for the hint. I fixed it.

PR #174, opened 2024-09-05
dropping legacy serialization support

FreeAndNil:
Fixes #175


ppkarwasz:
Could we just drop the `ISerializable` interface and `SerializableAttribute` altogether in 3.x for all runtimes?

Users that rely on that can use 2.x I guess.

FreeAndNil:
> Could we just drop the `ISerializable` interface and `SerializableAttribute` altogether in 3.x for all runtimes?
> 
> Users that rely on that can use 2.x I guess.
@ppkarwasz We would like to play safe as there was no feedback on #175.

.github/workflows/build.yaml
```diff
@@ -0,0 +1,51 @@
...
+      - name: Checkout repository
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332   # 4.1.7
+
+      - name: Set up dotnet
+        uses: actions/setup-dotnet@6bd8b7f7774af54e05809fcc5431931b3eb1ddee   # 4.0.1
+        with:
+          dotnet-version: 8
+          dotnet-quality: ga
+
+      - name: Build
```
fluffynuts, line 46: suggest: if you set up Node, you can do `npm test` to do all the build and test stuff you need via `zarro`. This way isn't wrong or anything, `zarro` is just more convenient, and will use the Quackers logger, which produces a useful final summary including failures and slow tests.

For instance, we do something like this on our builds:
```yml
    - name: ⚙️ Setup .NET
      uses: actions/setup-dotnet@v4
      with:
        dotnet-version: ${{ matrix.dotnet-version }}

    - name: 🐢 Setup NodeJs
      uses: actions/setup-node@v4
      with:
        node-version: ${{ matrix.node-version }}
        cache: 'npm'

    - name: 🏗️ Build and test 🧪
      env:
        DOTNET_SKIP_FIRST_TIME_EXPERIENCE: 1
        DOTNET_CLI_TELEMETRY_OPTOUT: 1
      run: |
        npm update -g npm --no-progress
        npm install --no-progress
        npm test
```

which runs all the tests with pretty, fully-namespaced output, and produces a summary like this:
image

as well as the list of slow tests and their timings above. What's considered slow? Default is any test taking more than 1s, but that can be guided by Quackers config (see: https://github.com/fluffynuts/quackers )

yes, quackers and zarro are more of my nonsense, so it's up to you. They are both currently installed - but I'd suggest updating them to latest 
- `npm install --save-dev zarro@latest`
- either upgrade Quackers.TestLogger via your preferred mechanism or use zarro, with the environment variable `UPGRADE_PACKAGES` (which can accept regex values) and `zarro upgrade-packages`. I can help with this if you like. Although it's a little slow (waiting on nuget.org a lot), it's hella convenient, and how I keep several of our more volatile packages up to date in our work projects.

If you decide to keep on using zarro and need any assistance, please shout.

fluffynuts (approved):

.github/workflows/build.yaml
ppkarwasz: @fluffynuts,

Feel free to modify the GitHub action. The current version is rather my attempt to understand how to build the project.

PR #181, opened 2024-09-16
automate version bumps

FreeAndNil:
Update the log4net version in every relevant file via script


FreeAndNil:
@fluffynuts they were already parameters, I just added default values for easier debugging.
I've removed the default values. Thanks for the hint.

Issue #189, opened 2024-09-26
separate review instructions for log4net


FreeAndNil:
- create a dedicated page in the Log4Net website
- replace the review instructions in the vote email template with a pointer to this new page
- revert the .NET-related changes in logging-parent

PR #190, opened 2024-09-26
#189 separate review instructions for log4net

FreeAndNil:
see https://logging.apache.org/log4net/release/release-review-instructions.html


fluffynuts (approved):

src/site/xdoc/release/release-review-instructions.xml
vy, line 1: Given Log4Net primarily targets Windows users, I think we should cater for them in the instructions. For instance, `umask`, `wget`, `cd $_`, `sha512sum`, `for sigFile ...`, etc. don't work in Windows. I suggest

- Separating Windows- and Linux-specific instructions
- Rewriting Windows-specific instructions

**An alternative route** can be to start with a warning telling Windows users to install bash/cygwin/subsystem/etc. to be able to run following Linux commands.

vy (changes requested):

src/site/xdoc/release/release-review-instructions.xml
FreeAndNil: @vy thanks for the hint.
I've added a PowerShell-Script for Windows users and updated the site accordingly.

doc/MailTemplate.txt
vy, line 1: You don't need `GitHub release (pre-release)` entry. Tags are mutable and you already shared the commit ID anyway.

vy (approved):
* I'm trusting you with `ps1` scripts, I haven't reviewed them.
* Dropped a minor remark about `MailTemplate.txt`.

doc/MailTemplate.txt
FreeAndNil: Thanks for the hint. I removed the line.

PR #196, opened 2024-10-17
more modern coding style

FreeAndNil:
- convert to filescoped namespaces
- use collection expressions
- switched to prefix `_` for private fields


src/log4net.Tests/Appender/AdoNet/Log4NetParameterCollection.cs
```diff
@@ -24,45 +24,44 @@
...
-    protected override void OnInsertComplete(int index, object? value)
+    if (value is IDataParameter param)
     {
-      base.OnInsertComplete(index, value);
-
-      if (value is IDataParameter param)
-      {
-        m_parameterNameToIndex[param.ParameterName] = index;
-      }
+      parameterNameToIndex[param.ParameterName] = index;
```
fluffynuts, line 39: I'd like to recommend standardising on the default C# style here - private fields should be prefixed with `_` - in particular, that makes it much easier (at least, for me) to determine _where_ a variable comes from - without the `_` prefix, I have to double-check every time whether this is a local or an instance variable.

fluffynuts (commented):
Just the one comment now - I support removing the C-style `m` prefix, but I'd like to retain the C#-style `_` prefix for private fields, if for no other reason than it makes it much easier to determine, at a glance, the difference between a local var and an instance field. Thoughts?

src/log4net.Tests/Appender/AdoNet/Log4NetParameterCollection.cs
FreeAndNil: I understand your reasoning.
At my company we had a similar discussion 15 years ago 😅
We decided against _.
I think it was mainly because there was only one C developer in the new C# team and the Pascal developers didn't like underscores.
The IDE will always show you whether it is a field or a variable.
We even have a CA rule which forces you to remove _ or m_ (that's what triggered my m_ changes).
But I acknowledge that in pull request it is not that easy (at least without using an IDE).

fluffynuts: yeah, even with an ide, one can't just read the code and know outright - one has to mouse over the token in the document

there are some other files that are using the `_` prefix and it's the recommended / default style according to Microsoft too ( https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/coding-style/identifier-names ) so it's most likely to be the code style enforced by most people if they were attempting to contribute.

fluffynuts:
> Alright. You convinced me. Would you like me to change it in this PR or in another one?

I don't mind - wherever is most convenient (:

FreeAndNil:
@fluffynuts I've adjusted the names with the help of the refactoring in rider and added the missing entries in .editorconfig.

fluffynuts (approved):

Issue #106, opened 2024-02-09
Enable automatic test runs on commit


fluffynuts:
It would be highly beneficial to the project to have automatic test runs (`npm test`) for every commit. To this end, a GitHub action that (eventually) invokes `npm test` would be fantastic.

ppkarwasz:
@fluffynuts,

I tried to add an NPM-based test run as you suggested in: https://github.com/apache/logging-log4net/blob/master/.github/workflows/build.yaml

However this fails on Ubuntu and MacOS (see [run 10825983032](https://github.com/apache/logging-log4net/actions/runs/10825982032), whereas the dotnet-based [run 10825714954](https://github.com/apache/logging-log4net/actions/runs/10825714954) doesn't. Could you look into that?

ppkarwasz:
@fluffynuts, the logs are available in the two runs I cited before:

- [run 10825983032](https://github.com/apache/logging-log4net/actions/runs/10825982032)
- [run 10825714954](https://github.com/apache/logging-log4net/actions/runs/10825714954)

I don't know which logs are relevant, so I didn't include any.

gdziadkiewicz:
@ppkarwasz @fluffynuts, are you still working on that? If not, can I try to make it work?

FreeAndNil:
@gdziadkiewicz you can try.
Piotr and Davyd don't have enough time.

ppkarwasz:
@gdziadkiewicz,

As Jan said, go ahead, we are swamped with other tasks.
IIRC Jan said that the Log4Net version does not need NPM, so maybe the workflow should just use `dotnet`.

fluffynuts:
I'd suggest using npm though - `zarro` does "all the right things" by convention when testing dotnet projects. `npm test` should _just work everywhere_.

PR #199, opened 2024-10-20
Fix empty string received by .NET 8 users on Linux on userName

gdziadkiewicz:
### Background
While working on getting log4net.Ext.Json package to work with log4net 3 I noticed a regression. With log4net 2.0.10  the test checking if the user name is as expected worked both during local run on my Windows machine and on the CI Linux agent used by Gitlab. After upgrading log4net to version 3.0.1 it stopped working on the CI and started reporting that the user name is an empty string.

Passing:
https://gitlab.com/gdziadkiewicz/log4net.Ext.Json/-/jobs/7522362244

Failing:
https://gitlab.com/gdziadkiewicz/log4net.Ext.Json/-/jobs/8094529032#L172

### Testing
It would be great to be able to run tests for log4net on all combinations of frameworks and OSes automatically (would you be interested in someone getting started work on it?).

In the meantime it draws as important for me to test (not sure about expected results yet):
* windows + net8 
* windows + net6
* windows + net461
* linux + net8
* linux + net6
* mac + net8
* mac + net6
* linux + mono
* mac + mono
* windows + mono

### Questions to answer before finalizing the change
1. Is WindowsIdentity Windows specific (name seems to suggest it)?
2. How is `Environment.User` implemented on different platforms? Does it change? Is it expensive? Would caching make sense?
3. Does and WindowsIdentity work on Mono on Win? If yes, how?
4. [RuntimeInformation.IsOSPlatform](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.runtimeinformation.isosplatform?view=netframework-4.7.1) is supported from net471, would reformulating the #if around it to communicate that help in the future?

### DIFF
--- a/src/log4net.Tests/Core/LoggingEventTest.cs
+++ b/src/log4net.Tests/Core/LoggingEventTest.cs
@@ -142,4 +142,15 @@ public sealed class LoggingEventTest
     static void AssertIsCurrentThreadId(string name)
       => Assert.That(SystemInfo.CurrentThreadId.ToString(CultureInfo.InvariantCulture), Is.EqualTo(name));
   }
+
+  [Test]
+  public void UserNameTest()
+  {
+    var expectedUserName =
+                Environment.OSVersion.VersionString.StartsWith("Microsoft Windows")?
+                  $"{Environment.UserDomainName}\\{Environment.UserName}"
+                : Environment.UserName;
+    var sut = new LoggingEvent();
+    Assert.AreEqual(expectedUserName, sut.UserName);
+  }
 }
--- a/src/log4net.Tests/Core/LoggingEventTest.cs
+++ b/src/log4net.Tests/Core/LoggingEventTest.cs
@@ -146,11 +146,11 @@ public sealed class LoggingEventTest
   [Test]
   public void UserNameTest()
   {
-    var expectedUserName =
+    string expectedUserName =
                 Environment.OSVersion.VersionString.StartsWith("Microsoft Windows")?
                   $"{Environment.UserDomainName}\\{Environment.UserName}"
                 : Environment.UserName;
-    var sut = new LoggingEvent();
-    Assert.AreEqual(expectedUserName, sut.UserName);
+    LoggingEvent sut = new();
+    Assert.That(sut.UserName, Is.EqualTo(expectedUserName));
   }
 }
--- a/src/log4net/Core/LoggingEvent.cs
+++ b/src/log4net/Core/LoggingEvent.cs
@@ -745,15 +745,27 @@ public class LoggingEvent : ILog4NetSerializable
   {
     if (_platformDoesNotSupportWindowsIdentity)
     {
-      // we've already received one PlatformNotSupportedException
+      // we've already received one PlatformNotSupportedException or null from TryReadWindowsIdentityUserName
       // and it's highly unlikely that will change
       return Environment.UserName;
     }
 
     try
     {
-      return _cachedWindowsIdentityUserName ??=
-          TryReadWindowsIdentityUserName();
+      if(_cachedWindowsIdentityUserName is not null)
+      {
+        return _cachedWindowsIdentityUserName;
+      }
+      else if(TryReadWindowsIdentityUserName() is string userName)
+      {
+        _cachedWindowsIdentityUserName = userName;
+        return _cachedWindowsIdentityUserName;
+      }
+      else
+      {
+        _platformDoesNotSupportWindowsIdentity = true;
+        return Environment.UserName;
+      }
     }
     catch (PlatformNotSupportedException)
     {
@@ -777,12 +789,17 @@ public class LoggingEvent : ILog4NetSerializable
   }
 
   private string? _cachedWindowsIdentityUserName;
-  private static string TryReadWindowsIdentityUserName()
+  
+  /// <summary>
+  /// TODO
+  /// </summary>
+  /// <returns>TODO</returns>
+  private static string? TryReadWindowsIdentityUserName()
   {
 #if !NET462_OR_GREATER
     if (!RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
     {
-      return string.Empty;
+      return null;
     }
 #endif
     using var identity = WindowsIdentity.GetCurrent();
--- a/src/log4net/Core/LoggingEvent.cs
+++ b/src/log4net/Core/LoggingEvent.cs
@@ -743,15 +743,15 @@ public class LoggingEvent : ILog4NetSerializable
 
   private string? TryGetCurrentUserName()
   {
-    if (_platformDoesNotSupportWindowsIdentity)
-    {
-      // we've already received one PlatformNotSupportedException or null from TryReadWindowsIdentityUserName
-      // and it's highly unlikely that will change
-      return Environment.UserName;
-    }
-
     try
     {
+      if (_platformDoesNotSupportWindowsIdentity)
+      {
+        // we've already received one PlatformNotSupportedException or null from TryReadWindowsIdentityUserName
+        // and it's highly unlikely that will change
+        return Environment.UserName;
+      }
+    
       if (_cachedWindowsIdentityUserName is not null)
       {
         return _cachedWindowsIdentityUserName;
@@ -788,14 +788,16 @@ public class LoggingEvent : ILog4NetSerializable
   private string? _cachedWindowsIdentityUserName;
   
   /// <returns>
-  ///  On Windows: UserName in case of success, empty string for null
+  ///  On Windows: UserName in case of success, empty string for unexpected null in identity or Name
   ///  <para/>
   ///  On other OSes: null
   /// </returns>
   /// <exception cref="PlatformNotSupportedException">Thrown on non-Windows platforms on net462</exception>
   private static string? TryReadWindowsIdentityUserName()
   {
-#if !NET462_OR_GREATER
+    // According to docs RuntimeInformation.IsOSPlatform is supported from netstandard1.1,
+    // but it's erroring in runtime on < net471
+#if NET471_OR_GREATER || NETSTANDARD2_0_OR_GREATER
     if (!RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
     {
       return null;
--- a/src/log4net/Core/LoggingEvent.cs
+++ b/src/log4net/Core/LoggingEvent.cs
@@ -701,8 +701,8 @@ public class LoggingEvent : ILog4NetSerializable
   /// </value>
   /// <remarks>
   /// <para>
-  /// Calls <c>WindowsIdentity.GetCurrent().Name</c> to get the name of
-  /// the current windows user.
+  /// On Windows it calls <c>WindowsIdentity.GetCurrent().Name</c> to get the name of
+  /// the current windows user. On other OSes it calls Environment.UserName.
   /// </para>
   /// <para>
   /// To improve performance, we could cache the string representation of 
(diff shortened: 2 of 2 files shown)


src/log4net/Core/LoggingEvent.cs
```diff
@@ -742,15 +742,27 @@ private static string ReviseThreadName(string? threadName)
...
+      // we've already received one PlatformNotSupportedException or null from TryReadWindowsIdentityUserName
       // and it's highly unlikely that will change
       return Environment.UserName;
     }
 
     try
     {
-      return _cachedWindowsIdentityUserName ??=
-          TryReadWindowsIdentityUserName();
+      if(_cachedWindowsIdentityUserName is not null)
```
FreeAndNil, line 752:
```suggestion
      if (_cachedWindowsIdentityUserName is not null)
```

src/log4net/Core/LoggingEvent.cs
```diff
@@ -742,15 +742,27 @@ private static string ReviseThreadName(string? threadName)
...
+      {
+        return _cachedWindowsIdentityUserName;
+      }
+      else if(TryReadWindowsIdentityUserName() is string userName)
```
FreeAndNil, line 756:
```suggestion
      if (TryReadWindowsIdentityUserName() is string userName)
```

src/log4net/Core/LoggingEvent.cs
```diff
@@ -742,15 +742,27 @@ private static string ReviseThreadName(string? threadName)
...
+      {
+        _cachedWindowsIdentityUserName = userName;
+        return _cachedWindowsIdentityUserName;
+      }
+      else
```
FreeAndNil, line 761:
```suggestion
```

src/log4net/Core/LoggingEvent.cs
```diff
@@ -742,15 +742,27 @@ private static string ReviseThreadName(string? threadName)
...
+      {
```
FreeAndNil, line 762:
```suggestion
```

src/log4net/Core/LoggingEvent.cs
```diff
@@ -742,15 +742,27 @@ private static string ReviseThreadName(string? threadName)
...
+        _platformDoesNotSupportWindowsIdentity = true;
```
FreeAndNil, line 763:
```suggestion
      _platformDoesNotSupportWindowsIdentity = true;
```

src/log4net/Core/LoggingEvent.cs
```diff
@@ -742,15 +742,27 @@ private static string ReviseThreadName(string? threadName)
...
+        return Environment.UserName;
```
FreeAndNil, line 764:
```suggestion
      return Environment.UserName;
```

src/log4net/Core/LoggingEvent.cs
```diff
@@ -742,15 +742,27 @@ private static string ReviseThreadName(string? threadName)
...
+      }
```
FreeAndNil, line 765:
```suggestion
```

src/log4net/Core/LoggingEvent.cs
```diff
@@ -742,15 +742,27 @@ private static string ReviseThreadName(string? threadName)
   {
     if (_platformDoesNotSupportWindowsIdentity)
     {
-      // we've already received one PlatformNotSupportedException
+      // we've already received one PlatformNotSupportedException or null from TryReadWindowsIdentityUserName
```
FreeAndNil, line 745: Can you fix the comment on UserName (especially clarify the "NOT AVAILABLE" part)?

FreeAndNil (commented):
added some hints

gdziadkiewicz:
@FreeAndNil Thanks for the extra quick check of my draft! I will iterate on the suggestions and questions/stuff in description. What is the deadline for 3.0.3?

FreeAndNil:
There is no deadline yet.
I think we will release in the middle of November, but this is negotiable 😅

FreeAndNil:
Adding some unit tests would be very nice.

src/log4net/Core/LoggingEvent.cs
gdziadkiewicz: I will fix it.

gdziadkiewicz:
I will add the unit tests.

FreeAndNil:
Hi @gdziadkiewicz,

trying to answer some of your questions.

> ### Testing
> 
> It would be great to be able to run tests for log4net on all combinations of frameworks and OSes automatically (would you be interested in someone getting started work on it?).

Yes, there is #106 and a github-actions branch from @rm5248 - I didn't find the time yet to take a closer look.

> In the meantime it draws as important for me to test (not sure about expected results yet):
> 
>     * windows + net8
>     * windows + net6
>     * windows + net461
>     * linux + net8
>     * linux + net6
>     * mac + net8
>     * mac + net6
>     * linux + mono
>     * mac + mono
>     * windows + mono

I think we can skip net6 (support ends in 3 weeks) and mono on windows (I don't think many users will do this).
> 
> ### Questions to answer before finalizing the change
> 
>     1. Is WindowsIdentity Windows specific (name seems to suggest it)?

Yes. https://github.com/dotnet/runtime/tree/main/src/libraries/System.Security.Principal.Windows/src
>     2. How is `Environment.User` implemented on different platforms? Does it change? Is it expensive? Would caching make sense?

It should work on all platforms, but it seems to give different results in web applications.
See https://stackoverflow.com/questions/33665929/whats-the-difference-between-system-environment-username-and-user-identity-name
>     3. Does and WindowsIdentity work on Mono on Win? If yes, how?

Yes. Same as net471. https://github.com/mono/mono/blob/main/mcs/class/corlib/System.Security.Principal/WindowsIdentity.cs
>     4. [RuntimeInformation.IsOSPlatform](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.runtimeinformation.isosplatform?view=netframework-4.7.1) is supported from net471, would reformulating the #if around it to communicate that help in the future?

Do you mean switching from net462 to net471 or adding a comment in the code?

Caching of the username seems not to be necessary. I get around 500 ms for 10.000 calls.

src/log4net/Core/LoggingEvent.cs
gdziadkiewicz: Done

src/log4net/Core/LoggingEvent.cs
gdziadkiewicz: Done

src/log4net/Core/LoggingEvent.cs
gdziadkiewicz: Done

src/log4net/Core/LoggingEvent.cs
gdziadkiewicz: Done

src/log4net/Core/LoggingEvent.cs
gdziadkiewicz: Done

gdziadkiewicz:
My plan was to find the answers alone, without using your time :D Thank you for the research!

> Do you mean switching from net462 to net471 or adding a comment in the code?

Current code guarding the code is `#if !NET462_OR_GREATER` which, combined with us targeting net462 and netstandard20,  means we won't include the call in any classic .net framework, only netstandard20. That works , but the way matching the docs would to only exclude it in net462 (cause that's where it is missing). Right now it doesn't matter, but if we retarget to net472 or something in the future that call which could be used there to check the OS easily will be excluded. So changing the #if would be future-proofing.

gdziadkiewicz:
I added the unit test and tested on Windows.

src/log4net/Core/LoggingEvent.cs
gdziadkiewicz: Done

src/log4net/Core/LoggingEvent.cs
gdziadkiewicz: Done

src/log4net/Core/LoggingEvent.cs
```diff
@@ -774,12 +783,17 @@ private static string ReviseThreadName(string? threadName)
   }
 
   private string? _cachedWindowsIdentityUserName;
-  private static string TryReadWindowsIdentityUserName()
+  
+  /// <summary>
+  /// TODO
```
gdziadkiewicz, line 788: Add

gdziadkiewicz:
@FreeAndNil I'm wondering about two parts of the code. One is the:
```csharp
    catch
    {
      return null;
    }
```
Should it really be the only place that doesn't fail back to `Environment.User`

and the second is the first if utilizing the flag:
```csharp
    if (_platformDoesNotSupportWindowsIdentity)
    {
      // we've already received one PlatformNotSupportedException or null from TryReadWindowsIdentityUserName
      // and it's highly unlikely that will change
      return Environment.UserName;
    }
```
 It is the only piece of code outside of the try block. It would be highly unfortunate to get an unhandled exception from it(based on impl details `Environment.UserName` can throw ). WDYT?

src/log4net.Tests/Core/LoggingEventTest.cs
```diff
@@ -138,4 +138,15 @@ static string ReviseThreadName(string? name)
     static void AssertIsCurrentThreadId(string name)
       => Assert.AreEqual(name, SystemInfo.CurrentThreadId.ToString(CultureInfo.InvariantCulture));
   }
+
+  [Test]
+  public void UserNameTest()
+  {
+    var expectedUserName =
```
FreeAndNil, line 145:
```suggestion
    string expectedUserName =
```
We switched away from var-Style some time ago.

src/log4net.Tests/Core/LoggingEventTest.cs
```diff
@@ -138,4 +138,15 @@ static string ReviseThreadName(string? name)
...
+                Environment.OSVersion.VersionString.StartsWith("Microsoft Windows")?
+                  $"{Environment.UserDomainName}\\{Environment.UserName}"
+                : Environment.UserName;
+    var sut = new LoggingEvent();
```
FreeAndNil, line 149:
```suggestion
    LoggingEvent sut = new();
```
is shorter ;-)

FreeAndNil:
> My plan was to find the answers alone, without using your time :D Thank you for the research!
> 
> > Do you mean switching from net462 to net471 or adding a comment in the code?
> 
> Current code guarding the code is `#if !NET462_OR_GREATER` which, combined with us targeting net462 and netstandard20, means we won't include the call in any classic .net framework, only netstandard20. That works , but the way matching the docs would to only exclude it in net462 (cause that's where it is missing). Right now it doesn't matter, but if we retarget to net472 or something in the future that call which could be used there to check the OS easily will be excluded. So changing the #if would be future-proofing.

Good point, you can adjust this.

FreeAndNil:
> @FreeAndNil I'm wondering about two parts of the code. One is the:
> 
> ```cs
>     catch
>     {
>       return null;
>     }
> ```
> 
> Should it really be the only place that doesn't fail back to `Environment.User`
> 

@fluffynuts you did the last meaningful change at this line in https://github.com/apache/logging-log4net/commit/e137855bdca5ed1a361c361846381607a5568d60 WDYT?

> and the second is the first if utilizing the flag:
> 
> ```cs
>     if (_platformDoesNotSupportWindowsIdentity)
>     {
>       // we've already received one PlatformNotSupportedException or null from TryReadWindowsIdentityUserName
>       // and it's highly unlikely that will change
>       return Environment.UserName;
>     }
> ```
> 
> It is the only piece of code outside of the try block. It would be highly unfortunate to get an unhandled exception from it(based on impl details `Environment.UserName` can throw ). WDYT?

We could catch errors and return NOTAVAILABLE.

BTW I recently activated the .net analyzers in #201 and changed all the catch-alls, so please merge master.

gdziadkiewicz:
Hi, I implemented the suggestions, extended the try, and refreshed the branch with Master. It's good to be reviewed now :)

FreeAndNil:
LGTM, thanks a lot.

PR #200, opened 2024-10-22
bumped NUnit to 4.2.2

FreeAndNil:
- added NUnit.Analyzers before moving to NUnit 4
- switched to NUnit 4.2.2

### DIFF
--- a/src/log4net.Tests/Util/CyclicBufferTest.cs
+++ b/src/log4net.Tests/Util/CyclicBufferTest.cs
@@ -27,14 +27,14 @@ using NUnit.Framework;
 namespace log4net.Tests.Util;
 
 /// <summary>
-/// Used for internal unit testing the <see cref="PropertiesDictionary"/> class.
+/// Used for internal unit testing the <see cref="CyclicBuffer"/> class.
 /// </summary>
 [TestFixture]
 public class CyclicBufferTest
 {
   [Test]
-  public void TestConstructorSize0() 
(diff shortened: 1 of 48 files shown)


src/log4net.Tests/Util/CyclicBufferTest.cs
```diff
@@ -33,49 +33,47 @@ namespace log4net.Tests.Util;
 public class CyclicBufferTest
 {
   [Test]
-  public void TestConstructorSize0()
-  {
-    Assert.Throws<ArgumentOutOfRangeException>(() => new CyclicBuffer(0));
-  }
+  public void TestConstructorSize0() 
+    => Assert.Throws<ArgumentOutOfRangeException>(() => new CyclicBuffer(0));
```
fluffynuts, line 37: I know that a lot of the Assert.* assertions were removed in favor of using `Assert.That` - there's an `Assert.That` for verifying that things do or don't throw as well, eg:

```csharp
            Assert.That(
                () => throw new InvalidOperationException("123"),
                Throws.Exception.InstanceOf<InvalidOperationException>()
                    .With.Message.EqualTo("123")
            );
```

it seems like `Assert.Throws` is still valid, but I wonder for how long?

fluffynuts (approved):
lgtm; just a suggestion around replacing `Assert.Throws`, but it's not imperative.

src/log4net.Tests/Util/CyclicBufferTest.cs
FreeAndNil: Thanks for the hint. I've adjusted the tests.

PR #201, opened 2024-10-23
Enable Microsoft.CodeAnalysis.NetAnalyzers

FreeAndNil:
activated Microsoft.CodeAnalysis.NetAnalyzers and fixed Warnings

### DIFF
--- a/src/log4net.Tests/Appender/AdoNet/Log4NetCommand.cs
+++ b/src/log4net.Tests/Appender/AdoNet/Log4NetCommand.cs
@@ -24,7 +24,7 @@ using System.Data;
 
 namespace log4net.Tests.Appender.AdoNet;
 
-public class Log4NetCommand : IDbCommand
+internal sealed class Log4NetCommand : IDbCommand
 {
   public Log4NetCommand()
   {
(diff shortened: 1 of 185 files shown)


fluffynuts (approved):
lgtm

src/log4net.Tests/Appender/AdoNet/Log4NetCommand.cs
```diff
@@ -24,7 +24,7 @@
 
 namespace log4net.Tests.Appender.AdoNet;
 
-public class Log4NetCommand : IDbCommand
+internal sealed class Log4NetCommand : IDbCommand
```
ppkarwasz, line 27: Just a curiosity: isn't this a breaking change?
At least it would be a breaking change in Java.

FreeAndNil: Yes, but this is in our test assembly and we don't ship it ;-)
This class should better be named MockDbCommand - I will change it in my next PR.

src/log4net/ObjectRenderer/RendererMap.cs
```diff
@@ -217,7 +211,7 @@ public IObjectRenderer Get(Type type)
   /// <summary>
   /// Gets the default renderer instance
   /// </summary>
-  public IObjectRenderer DefaultRenderer => _sDefaultRenderer;
+  public static IObjectRenderer DefaultRenderer { get; } = new DefaultRenderer();
```
gdziadkiewicz, line 214: For record: this change in public API broke log4net 3.0.3 for log4net.Ext.Json 3.0.1.

FreeAndNil: reason for #211

PR #203, opened 2024-10-31
use raw strings for config files in unit tests

FreeAndNil:
- use raw strings for config files in unit tests
- fixed more warnings

### DIFF
--- a/src/log4net.Tests/Appender/AdoNetAppenderTest.cs
+++ b/src/log4net.Tests/Appender/AdoNetAppenderTest.cs
@@ -87,7 +87,8 @@ public class AdoNetAppenderTest
   public void WebsiteExample()
   {
     XmlDocument log4NetConfig = new();
-    log4NetConfig.LoadXml("""
+    log4NetConfig.LoadXml(
+      """
       <log4net>
       <appender name="AdoNetAppender" type="log4net.Appender.AdoNetAppender">
           <bufferSize value="-1" />
@@ -143,7 +144,7 @@ public class AdoNetAppenderTest
           <appender-ref ref="AdoNetAppender" />
         </root>  
       </log4net>
-    """);
+      """);
 
     ILoggerRepository rep = LogManager.CreateRepository(Guid.NewGuid().ToString());
     XmlConfigurator.Configure(rep, log4NetConfig["log4net"]!);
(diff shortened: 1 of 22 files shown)


src/log4net.Tests/Appender/RollingFileAppenderTest.cs
```diff
@@ -524,23 +523,10 @@ private static string GetCurrentFile() =>
...
-      if (sFiles[i].Trim().Length > 0)
-      {
-        result.Add(new RollFileEntry(sFiles[i], iBackupFileLength));
-      }
-    }
-
-    return result;
-  }
+  private static List<RollFileEntry> MakeBackupFileEntriesFromBackupGroup(string sBackupGroup, int iBackupFileLength) 
+    => sBackupGroup.Split(' ').Where(file => file.Trim().Length > 0)
```
fluffynuts, line 527: personal note: i really don't like inline method bodies - it's like I take a little longer to grok them over, eg

```csharp
private static List<RollFileEntry> MakeBackupFileEntriesFromBackupGroup(
    string sBackupGroup,
    int iBackupFileLength
)
{
    return sBackupGroup.Split(' ')
                                      .Where(file => file.Trim().Length > 0)
                                      .Select(file => new RollFileEntry(file, iBackupFileLength))
                                      .ToList()
}
```

Perhaps it's just me though...

I'd also like to make two "soft" suggestions for code style going forward:
1. vertical parameters (as in my example above) - see https://youtu.be/1WL-SudgJAo (I had to re-upload this because it had been removed, so I got it from archive.org and re-upped)
2. dropping type prefixes for variables / parameters (eg _i_BackupFileLength and _s_BackupGroup)

I wouldn't go on a major refactoring spree, just fix things as I'm in the area.

fluffynuts (approved):

FreeAndNil:
Good points. I will adjust the code.

src/log4net.Tests/Appender/AdoNetAppenderTest.cs
```diff
@@ -80,69 +80,70 @@ public void BufferingTest()
...
-                        <size value=""2000"" />
-                        <layout type=""log4net.Layout.ExceptionLayout"" />
-                    </parameter>
-                </appender>
-                <root>
-                    <level value=""ALL"" />
-                    <appender-ref ref=""AdoNetAppender"" />
-                  </root>  
-                </log4net>");
+    log4NetConfig.LoadXml("""
```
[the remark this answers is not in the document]
FreeAndNil: fixed

src/log4net.Tests/Appender/RollingFileAppenderTest.cs
FreeAndNil: fixed - the type prefixes for variables were only in RollingFileAppenderTest

PR #217, opened 2024-12-18
fixed SetQWForFiles method name in FileAppender from #196

### LINKED ISSUE #216: Breaking change to FileAppender in a third-digit release
A change to `FileAppender` that breaks backward compatibility was introduced with #196.
`SetQWForFiles` is a publicly visible method that was renamed to `SetQwForFiles`. Need to rename it back in order to keep types that extend `FileAppender` compatible with the latest version of log4net

sketchturner:
A change to `FileAppender` that breaks backward compatibility was introduced with #196.
Renamed `SetQwForFiles` back to `SetQWForFiles` in order to restore compatibility with types that extend `FileAppender`
Fixes #216

### DIFF
--- a/src/log4net/Appender/FileAppender.cs
+++ b/src/log4net/Appender/FileAppender.cs
@@ -1232,7 +1232,7 @@ public class FileAppender : TextWriterAppender
         _stream.AcquireLock();
         try
         {
-          SetQwForFiles(_stream);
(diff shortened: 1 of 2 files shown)


FreeAndNil:
@sketchturner 3.0.4 is out - https://github.com/apache/logging-log4net/releases/tag/rel%2F3.0.4

PR #222, opened 2025-01-30
Make copy of appender list when removing all, for increased thread-safety

tmct:
Hello,

We have been witnessing this error now and again while using log4net 2.0.15:

```
Collection was modified; enumeration operation may not execute.
   at log4net.Appender.AppenderCollection.Enumerator.MoveNext()
   at log4net.Util.AppenderAttachedImpl.RemoveAllAppenders()
   at log4net.Repository.Hierarchy.Logger.RemoveAllAppenders()
   at log4net.Repository.Hierarchy.Hierarchy.Shutdown()
   at log4net.Core.LoggerManager.Shutdown()
   at log4net.LogManager.Shutdown()
   ```
It looks like the code on master is still susceptible to this, what do you think of this change? (The try-catch should be wel-equipped to handle appenders that have left the List, so long as we can still access their name for the error message...)

Thanks,
Tom

### DIFF
--- a/src/log4net/Util/AppenderAttachedImpl.cs
+++ b/src/log4net/Util/AppenderAttachedImpl.cs
@@ -246,7 +246,7 @@ public class AppenderAttachedImpl : IAppenderAttachable
   {
     if (_appenderList is not null)
     {
-      foreach (IAppender appender in _appenderList)
+      foreach (IAppender appender in _appenderList.ToArray())
       {
         try
         {
--- a/src/log4net/Util/AppenderAttachedImpl.cs
+++ b/src/log4net/Util/AppenderAttachedImpl.cs
@@ -71,9 +71,7 @@ public class AppenderAttachedImpl : IAppenderAttachable
       return 0;
     }
 
-    _appenderArray ??= _appenderList.ToArray();
-
-    foreach (IAppender appender in _appenderArray)
+    foreach (IAppender appender in _appenderList.ToArray())
     {
       try
       {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by FreeAndNil** [comment]
@tmct thanks for your pull request.

I would prefer the pattern already used in AppendLoopOnAppenders():

```csharp
_appenderArray ??= _appenderList.ToArray();
foreach (IAppender appender in _appenderArray)
```

Can you adjust the code?

tmct:
Yes of course, consistency can be very useful! Thank you.

tmct:
Hi @FreeAndNil - I had a look and I don't think the private `_appenderArray` was actually doing any caching at all beyond the scope of each individual method. And it likely hinders the thread-safeness of anything that uses it. Most likely the original intent was to protect against exactly the thing we're seeing here.

I've removed it everywhere if that's ok?

FreeAndNil:
@tmct the _appenderArray had the advantage to copy the contents only once into an array.
Now every method call (even those who only read) create a new Array.

I'm currently considering switching to System.ComponentModel.Collections.Collection<IAppender> as base class.
But that would probably be a breaking change.

tmct:
Ah yes you could be right, this could specifically have a performance impact in AppendLoopOnAppenders, being the method among these that we expect to be called frequently without a change in the list, the problem being the repeated filing of the same array. 

I can address that later today or Monday, thanks.

FreeAndNil:
@tmct Will you find the time to complete within this week?

FreeAndNil:
Hi @tmct,

hope you're doing well! Just checking in on this.

Do you have an estimated timeframe for when this will be completed?
Can we agree to finish this PR by the end of next week?
It would be great to get this wrapped up soon.

Thanks.

FreeAndNil:
@tmct closing this for now, please reopen when you find time to complete.

PR #228, opened 2025-02-06
Show config file path in error message

### LINKED ISSUE #227: Provide path of expected .config file in error message
With log4net version 3.0.3:
I got the message:
"log4net:ERROR Failed to find configuration section 'log4net' in the **application's .config file**. Check your **.config** file for the <log4net> and <configSections> elements. The configuration section should look like: \<section name="log4net" type="log4net.Config.Log4NetConfigurationSectionHandler,log4net" />"

It would be helpful to have the name + absolute path of **"application's .config file"** which is used by log4net, in the above error message of log4net.

Background:
When I'm running my unit tests with "dotnet test ..." I got the above error message and I analyzed and wonder, what's wrong with my _AssemblyName.Test.dll.config_.
I had to spend some effort and used then
    `log4net.Util.LogLog.InternalDebugging = true;`
to find out, that when running 
`"dotnet test ..."` for TFM = net8.0, the expected config file was "**testhost.dll.config**"!
If this information would be displayed in the above error message, I (and possibly other users) could save a lot of time.

FreeAndNil:
fixes #227

### DIFF
--- a/src/log4net.Tests/Config/XmlConfiguratorTest.cs
+++ b/src/log4net.Tests/Config/XmlConfiguratorTest.cs
@@ -1,0 +1,60 @@
+#region Apache License
+//
+// Licensed to the Apache Software Foundation (ASF) under one or more 
+// contributor license agreements. See the NOTICE file distributed with
+// this work for additional information regarding copyright ownership. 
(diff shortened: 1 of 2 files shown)


fluffynuts (approved):
lgtm

PR #237, opened 2025-03-28
make RollingFileAppender.RollFile virtual

### LINKED ISSUE #234: Compressing rolled files
I'm really struggling to find a way to zip the rolled old `fromFile`, at the moment that log4net decides to rollover on date. I don't see any direct way to get notified if a log file was rolled, only indirectly and incomplete.

My kind request is to implement **ALT-3**, if possible.

### ALT-1: override RollOverRenameFiles() method

```
protected override void RollOverRenameFiles(string baseFileName)
{
    base.RollOverRenameFiles(baseFileName);
    //do zip
}
```

Result= NOK
-only works for rollover based on size
-for rollover based on date, this method is not called, no zip

### ALT-2: override AdjustFileBeforeAppend() method

```
protected override void AdjustFileBeforeAppend()
{
   var currentFile = File;
   base.AdjustFileBeforeAppend();
   if (!currentFile.Equals(File))
   {
      //do zip
   }
}
```

Result= NOK
-works for rollover based on size on same day
-when crossing a date boundary, `currentFile` differs from `File` => zip OK
-when starting the main app the next day, it misses the previous day file => no zip

### ALT-3: override RollFile(string fromFile, string toFile) method

```
protected override void RollFile(string fromFile, string toFile)
{
   base.RollFile(string fromFile, string toFile);
   //do zip on 'fromFile'
}
```

Cannot override now because method is not `virtual`

So, my request is: could this `RollFile()` be made virtual?

Thanks!

Greetings,
Liviu

FreeAndNil:
fixes #234

### DIFF
--- a/src/log4net/Appender/RollingFileAppender.cs
+++ b/src/log4net/Appender/RollingFileAppender.cs
@@ -1112,7 +1112,7 @@ public partial class RollingFileAppender : FileAppender
   /// also checks for existence of target file and deletes if it does.
   /// </para>
   /// </remarks>
-  protected void RollFile(string fromFile, string toFile)
+  protected virtual void RollFile(string fromFile, string toFile)
   {
     if (FileExists(fromFile))
     {


FreeAndNil:
See https://github.com/apache/logging-log4net/issues/234#issuecomment-2764652848

PR #246, opened 2025-05-19
#246 fix InvalidCastException in ThreadContextStack.InternalStack.set

### LINKED ISSUE #244: v3.1.0 - Exception with RollingFileAppender when files are already present
I just updated from v3.0.4 to 3.1.0 and suddenly logging fails with an ArgumentException ("The path is not of a legal form.")

Stack:
```
at System.IO.Path.NormalizePath(String path, Boolean fullCheck, Int32 maxPathLength, Boolean expandShortPaths)
   at System.IO.Path.InternalGetDirectoryName(String path)
   at log4net.Appender.RollingFileAppender.InitializeFromOneFile(String baseFile, String curFileName)
   at log4net.Appender.RollingFileAppender.InitializeRollBackups(String baseFile, IList`1 arrayFiles)
   at log4net.Appender.RollingFileAppender.DetermineCurSizeRollBackups()
   at log4net.Appender.RollingFileAppender.ExistingInit()
   at log4net.Appender.RollingFileAppender.ActivateOptions()
   at log4net.Program.Main(String[] args) in D:\Temp\Egbert\log4net\log4net\Program.cs:line 29
```

It turns out that when there's already a file present in the target directory, above mentioned error is thrown.

Below a repro - first time you run it, it will work because no file is present. The second time it fails:

```
static void Main(string[] args)
{
    PatternLayout patternLayout = new PatternLayout
    {
        ConversionPattern = @"%date [%thread] %-5level %logger [%property{NDC}] - %message%newline"
    };
    patternLayout.ActivateOptions();

    RollingFileAppender rollingFileAppender = new RollingFileAppender
    {
        AppendToFile = true,
        File = @"C:\TempLog\",
        Layout = patternLayout,
        DatePattern = "yyyy-MM-dd'.log'",
        MaxSizeRollBackups = 10,
        MaximumFileSize = "5MB",
        RollingStyle = RollingFileAppender.RollingMode.Composite,
        StaticLogFileName = false,
    };
    rollingFileAppender.ActivateOptions();
}
```

FreeAndNil:
fixes #244

### DIFF
--- a/src/log4net/Util/ThreadContextStack.cs
+++ b/src/log4net/Util/ThreadContextStack.cs
@@ -188,10 +188,10 @@ public sealed class ThreadContextStack : IFixingRequired
     set
     {
       _stack.Clear();
-      var frames = (StackFrame[])value.ToArray();
+      object[] frames = value.ToArray();
       for (int i = frames.Length - 1; i >= 0; i--)
       {
-        _stack.Push(frames[i]);
+        _stack.Push(frames[i].EnsureIs<StackFrame>());
       }
     }
   }
--- a/src/changelog/3.1.1/.release-notes.adoc.ftl
+++ b/src/changelog/3.1.1/.release-notes.adoc.ftl
@@ -1,0 +1,41 @@
+////
+    Licensed to the Apache Software Foundation (ASF) under one or more
+    contributor license agreements.  See the NOTICE file distributed with
+    this work for additional information regarding copyright ownership.
+    The ASF licenses this file to You under the Apache License, Version 2.0
+    (the "License"); you may not use this file except in compliance with
+    the License.  You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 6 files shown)


src/log4net/Util/ThreadContextStack.cs
```diff
@@ -188,10 +188,10 @@ internal Stack InternalStack
     set
     {
       _stack.Clear();
-      var frames = (StackFrame[])value.ToArray();
+      object[] frames = value.ToArray();
       for (int i = frames.Length - 1; i >= 0; i--)
       {
-        _stack.Push(frames[i]);
+        _stack.Push(frames[i].EnsureIs<StackFrame>());
```
fluffynuts, line 194: minor: I'd suggest rather using pettern-matching instead of throwing if something in that array is wrong, eg

```csharp
if (frames[i] is StackFrame s)
{
  _stack.Push(s)
}
```
This would stop a throw on invalid data set, but leave the collection empty, which may not be better - this may just be a stylistic choice on my part.

fluffynuts (approved):

PR #253, opened 2025-06-10
Asynchronous Sending for RemoteSyslogAppender in log4net

yogitasingh001:
This PR addresses a major performance bottleneck in RemoteSyslogAppender, where synchronous .Wait() calls were blocking high-throughput applications.

Summary of Changes
Introduced a background task and queue (producer-consumer model) for async UDP sends
Removed thread-blocking .Wait() from Append() method
Ensured graceful shutdown with cancellation tokens
Documentation included at docs/AsynchronousSendingforRemoteSyslogAppenderinlog4net.pdf

Request
Please consider reviewing and including this change in a future release. We rely on log4net and would prefer not to maintain a separate fork.


FreeAndNil:
@yogitasingh001 thanks for your proposal.
I will take a deeper look this week.

yogitasingh001:
@FreeAndNil thanks

FreeAndNil:
implements #255 
@yogitasingh001 I've converted your pdf into an issue.

PR #257, opened 2025-06-15
Add tests for regression in file rolling and fix bug

### LINKED ISSUE #250: v3.1.0 is breaking rolling logfiles usage
Starting with v3.1.0 my logfiles are no longer correctly rolled. It is only rolling max 1 additional file and gots constantly overwritten instead of having max 100 as configured. I use the following configuration:

    <appender name="Log"
              type="log4net.Appender.RollingFileAppender">

        <file type="log4net.Util.PatternString">
            <conversionPattern value="%property{LogFolder}\\Logging.log"/>
        </file>

        <param name="StaticLogFileName" value="true"/>
        <param name="AppendToFile" value="false"/>
        <rollingStyle value="Size"/>
        <maximumFileSize value="500MB"/>
        <maxSizeRollBackups value="100"/>
        <CountDirection value="1"/>
        <preserveLogFileNameExtension value="true"/>
        <lockingModel type="log4net.Appender.FileAppender+MinimalLock"/>

        <layout type="log4net.Layout.PatternLayout">
            <param name="ConversionPattern"
                   value="%date [%3thread] %-5level %logger{1} %-20message %n"/>
        </layout>

         ... 
    </appender>

To me it looks like related to the recent changes in issue report #232

gdziadkiewicz:
Fixes #250 

### Problem
It is easy to break someone's legitimate Log4net configuration or not account for .NET version quirks while making changes to the rolling appended.

### Idea
Gather all of the configurations we gathered in bug reports, plus variations that make sense for us, and use them as test cases against regression.

Remember:
* to run on both .NET and .NET Framework (this is free, we are running on net8 and net462 :) )
* to include both approaches of providing file extension -> in the file and the date field ( done thanks to the no_name test)
* to test rolling during one "long" run (added one with timeout)
* to test behaviour on restart ( #250 manifests during restart) (done)
* add comments explaining used configs

TODO:
 * [x] Fix paths on Linux and macOS
 * [x] Add comments
 * [x] Cleanup style


src/log4net/Appender/RollingFileAppender.cs
```diff
@@ -681,18 +681,16 @@ protected List<string> GetExistingFiles(string baseFilePath)
...
     {
       string fullPath = Path.GetFullPath(baseFilePath);
-      string dir = Path.GetDirectoryName(baseFilePath);
 
       directory = Path.GetDirectoryName(fullPath);
       if (Directory.Exists(directory))
       {
         string baseFileName = Path.GetFileName(fullPath);
-
+
```
Flohack74, line 689: You might want to clean that whitespace change.

gdziadkiewicz: Will do, thanks!

gdziadkiewicz:
I expect it to be review-ready today.

PR #294, opened 2026-05-08
Fix race in Hierarchy.TryCreateLogger

### LINKED ISSUE #292: Log messages not getting to appender during concurrent GetLogger
Hi
I'm looking into going to log4net 3.3 but we have some logging concurrency test that bails out and I'm uncertain about the real cause.

What happens is that we don't get expected log count in the appender when doing concurrent logging.

Rather hard to reproduce, see attached repro which sets up an ILoggerRepository and Hierarchy and then runs the concurrent logging.
(put them on disk and run script.ps1), (all of this on windows 11 and windows server)

The result I get:
```
3.3.0  on net472:           OK=17, NOK=3
3.3.0  on net10.0-windows:  OK=19, NOK=1
3.0.4  on net472:           OK=13, NOK=7
3.0.4  on net10.0-windows:  OK=20, NOK=0
2.0.14 on net472:           OK=20, NOK=0
2.0.14 on net10.0-windows:  OK=20, NOK=0
2.0.17 on net472:           OK=20, NOK=0
2.0.17 on net10.0-windows:  OK=20, NOK=0
(not ok should be 0)
```
Please have a look at the LoggingConcurrencyTests ctor if that code is correct for log4net 2 and 3 versions.
Please try to reproduce it with attached code.

The actual test just does massive parallel logging, including a GetLogger call.

If the GetLogger call is done outside the loop all is good ('alt b') so it's related to the GetLogger.

It's possible to argue that the test is somewhat theoretic and maybe aggressive but I think that this problem seen on v3, not v2, is a bit worrying.

(hmm, working on getting the repro code into this..)

gdziadkiewicz:
This PR adds the regression (compared to log4net version 2) test case from #292 and fix in form of bringing back lock in Hierarchy.TryCreateLogger that is used to synchronize creation and registering of new Loggers. Prior to the change it was possible to get an unregistered logger that will result in dropped messages.

Resolves #292


gdziadkiewicz:
Two of the new tests are on the [slow list](https://github.com/apache/logging-log4net/actions/runs/25543175476/job/74973336778?pr=294#step:5:558)  :
1. log4net.Tests.Hierarchy.LoggingConcurrencyTest.GetLogger_UnderHighConcurrency_ShouldNotLoseMessages **(2.56 s)**
2. log4net.Tests.Hierarchy.LoggingConcurrencyTest.GetLogger_WithDottedNamesUnderHighConcurrency_ShouldNotLoseMessages **(2.13 s)**

FreeAndNil (approved):
Thanks a lot.

PR #296, opened 2026-05-19
extract SECURITY.md and add a remediation time frame policy

FreeAndNil:
extract SECURITY.md and add a remediation time frame policy


FreeAndNil:
no longer considered useful

PR #302, opened 2026-07-28
log4net.Ext.Mail

### LINKED ISSUE #300: SmtpAppender uses not recomended SmptClient
We use the SmtpAppender in a number of projects.

When used with modern mail servers, connection issues are becoming increasingly common.

Until now, the SmtpClient class has been used for sending email, but Microsoft no longer recommends its use:

> _Important
> We don't recommend that you use the SmtpClient class for new development because SmtpClient doesn't support many modern protocols. Use MailKit or other libraries instead. For more information, see SmtpClient shouldn't be used._
[https://learn.microsoft.com/en-us/dotnet/api/system.net.mail.smtpclient?view=net-10.0#remarks](https://learn.microsoft.com/en-us/dotnet/api/system.net.mail.smtpclient?view=net-10.0#remarks)

Is the project planning to make the recommended switch to MailKit in the near future?

FreeAndNil:
fixes #300

Microsoft no longer recommends `System.Net.Mail.SmtpClient`, and it increasingly
fails against modern mail servers. This adds a satellite assembly
`log4net.Ext.Mail` with a MailKit based `SmtpAppender`, so log4net itself takes on
no mail dependency.

- Same options as `log4net.Appender.SmtpAppender` - migrating means changing only
  the `type` attribute.
- Behaviour differences: `smtpHost` is required, `enableSsl` negotiates implicit
  TLS/STARTTLS automatically, `Ntlm` needs explicit credentials (MailKit cannot
  reuse the Windows logon session), and semicolon-separated recipient lists now
  parse correctly.
- Site docs now recommend the new appender; all SMTP samples use it.
- Adds `Log4NetAssert.EnsureNotNullOrEmpty`.

Also included, unrelated to #300: TelnetAppender registers the client before
sending the welcome banner, fixing a `HasConnections` race that dropped messages.


fluffynuts (approved):
lgtm

PR #315, opened 2026-09-01
Escaping special characters in multiple appenders

FreeAndNil:
Follow-ups from the second security scan, stacked on #314. One rule, applied to
every sink that had got it wrong: content a sink cannot carry is escaped visibly,
never silently deleted, and never allowed to cost the record or the event.

- `LocalSyslogAppender` escapes newlines, which reached `syslog(3)` unchanged and
  let content forge a second record. `NewLineHandling` mirrors the remote appender,
  and `SyslogNewLineHandling` moves out of `RemoteSyslogAppender` so both can use
  it. Measured: glibc does no escaping of its own, contrary to the old comment.
- `EventLogAppender` and `OutputDebugStringAppender` escape NUL, which ended the
  stored record with no error. Measured on Windows: a 45 character message with a
  NUL at 23 stored as its 23 character prefix.
- `RemoteSyslogAppender` escapes characters outside RFC 3164 rather than deleting
  them. "Schönwetter 你好" reached the collector as "Schnwetter ".
- `TelnetAppender` and `SmtpPickupDirAppender` escape unpaired surrogates. Both
  used a writer whose encoding throws: the first disconnected every connected
  client, the second destroyed the whole buffered batch and left a truncated mail.
- `EventLogAppender`'s size limit is computed rather than guessed. The old constant
  sat above the point where the service silently discards the record, so log4net
  truncated events into oblivion.
 
Behaviour changes: syslog exception traces are now one escaped record instead of
several lines, `Keep` restores the old behaviour; `RemoteSyslogAppender.SyslogNewLineHandling` is no longer nested,
configuration binds by value name and is unaffected.


gdziadkiewicz (approved):
LGTM

